PR #6, opened 2017-03-08
Update the zoneinfo to IANA 2017b

ehoogerbeets:


gouniLee:
I have a question. Some of "f" values in zoneinfo are changes as number such as +03 or +08/+09 etc.. . and some files are still old values such as KST. A "f" value is inconsistent.  is it correct result?

for example,
Previously, Asia/Vietiane.json was written as below.
`{
    "f": "ICT",
    .....
}`

But now it changes as below. Asia/Vietiane.json
`{
    "f": "+07",
    .....
}`

Some files are having still previous values. Asia/Seoul.json
`{
       "f": "KST",
        ......
}`

PR #16, opened 2017-05-16
Fix unit tests - various unit test fixes for Nodejs, Ubuntu, and Mac

ehoogerbeets:


js/lib/ilib.js
```diff
@@ -129,7 +129,7 @@ ilib._getPlatform = function () {
     		}
     	} catch (e) {}
     	
-        if (typeof(process) !== 'undefined' && typeof(window) === 'undefined' && typeof(module) !== 'undefined') {
```
gouniLee: Is there specific reason to rollback as previous code? This changes from Enact (Enyo+React) teams's report.
fyi.
https://github.com/enyojs/iLib/pull/94

ehoogerbeets: Yes, the tests with the code as it was for Enact did not work with the regular nodejs tests without Enact. The reason is that the tests are run inside of a nodejs "vm" which has a "window" variable in its context. (See https://nodejs.org/api/vm.html for reference documentation.) Many tests failed because of this, even though the code itself was fine. Ilib needs to find some other way of detecting Enact as a platform.

gouniLee: When I ran a unittest on nodejs, I didn't get any fail cases in regards to that change. Could you tell me a your test environment?

ehoogerbeets: It was ubuntu with the full unit tests (including legacy) on nodejs 6.

gouniLee:
Seonghyup ran a unittest with this PR on Windows7 with node version 4.0.0 and 6.10.3 . I confirmed all of unittest is passed.

ehoogerbeets:
This PR needs a github approval before it can be merged. Please approve it.

PR #24, opened 2017-05-29
Changes to get Travis-CI working

ehoogerbeets:


ehoogerbeets:
Travis-CI now gets all the way through the ant build steps and runs the unit tests. Unfortunately, they do not all pass, but at least travis is working correctly now.

PR #29, opened 2017-06-09
Fix unit tests

ehoogerbeets:


ehoogerbeets:
Changes are merged into the "makeTravisWork" branch because otherwise it would not pass. So that is a joint PR, and this PR can be abandoned.

PR #30, opened 2017-06-10
Do what is necessary to get travis working in master

ehoogerbeets:
Merged over the stuff in development branch that made it work properly there.


gouniLee:
Confirmed. Not breaking unittest on dynamic version on node and assembled with jsunit.
But I noticed root/testSuiteDyn.html had some failures. Could you check on this?

PR #36, opened 2017-06-16
Implement format() for LisFormatter and Add more testcases

gouniLee:
Implement format() for ListFormatter.
Add more testcases for verify APIs.


js/lib/ListFmt.js
```diff
@@ -118,8 +131,39 @@ var ListFmt = function(options) {
  * @returns {String} a string containing the list of items that
  * is grammatically correct for the locale of this formatter
  */
+
 ListFmt.prototype.format = function(items) {
+	var itemCount = items.length;
```
ehoogerbeets: Need some safety code here in case "items" is undefined or not an array. Also need to add some unit tests to verify that the checks are working correctly.

gouniLee:
Following your review, I add safety code and more test cases. also fixed conflict code.

PR #39, opened 2017-07-05
Fix bug of time formatter

Seonghyup:
When formatting 'mm:s' template(especially, via duration formatter using clock style), the time is formatted using minute, not second.
I fixed to use second for fomatting 'mm:s' template, and add unit test case for this.


ehoogerbeets:
It looks like this fix may have broken some other tests. I suspect the tests need to be updated. Please have a look at those to see if they are related before you merge this.

Seonghyup:
You're right. I've missed wrong test cases existing. Thank you.

PR #42, opened 2017-07-14
Name formatting updates

ehoogerbeets:
- Support new "honorific" field which can be a prefix or suffix depending on the locale
- Support "formal_short" and "formal_long" format styles which allow you to format names like "Dr. Smith" and "Dr. Alicia Smith". That is, honorific + family or honorific + given + family names.
- Update Korean name formatting so that prefixes and suffixes are both separated from the name with a space
- Updated documentation to make it more clear


Seonghyup:
Looks good to me!

PR #46, opened 2017-07-24
Expand nodeunit test coverage and make it the default test suite

ehoogerbeets:
- added support for assembled code + dynamic data (test.dyndata.uncompiled.nu and test.dyndata.compiled.nu)
- added support for assembled code + assembled data (test.assembled.uncompiled.nu and test.assembled.compiled.nu)
- already had support for dynamic code + dynamic data before (test.dynamic.uncompiled.nu and test.dynamic.compiled.nu)
- fixed up the dependencies for all these targets
- added "test.<suite>.nu" targets to run only that suite in dynamic code&data mode 
- added "debug.<suite>.nu" targets to run only that suite in the remote debugger (chrome). Cut-and-paste the URL it gives you into chrome to begin debugging that suite.
- you can run the suite in a browser by running "ant.remote" at the top level, and then going to a browser and entering "http://localhost:9090/test/<suite>/nodeunit/testSuite.html" for any particular suite
-  you can run all the tests in a browser by running "ant.remote" at the top level, and then going to a browser and entering "http://localhost:9090/test/testSuite.html"
- switched "ant test" to use the node unit tests instead of jsunit
- the jsunit tests are still there and available if necessary


gouniLee:
I thinks you missed to commit file nodeunit/testSuite.html for running a whole  bunch cases at a one time. "http://localhost:9090/test/testSuite.html" is not working.  Could you check this?

ehoogerbeets:
The testSuite.html is added now

PR #54, opened 2017-10-02
Emergency numbers support

ehoogerbeets:


ehoogerbeets:
Yes, it is only json files for now. In HealthTap, we needed these numbers in the Ruby code. I have not written the js code yet, as it depended on google location APIs to locate the user's country and from there determine their local emergency number. The google location API is not free. HT has a subscription, so we can use it, but open-source people cannot. So, I just contributed the portion that I could, which was the data.

PR #60, opened 2017-11-10
Update locmaker.js.output. ES6 'import' cannot be used at runtime

gouniLee:
A locmaker.js code should be more generic. But I can't update at one time for working properly. I think it better to be done step by step. 

Here's an issue that addressed from @JayCanuck

Currently, The output of locmaker tool code, insert the at the first line as below:
`import * as ilib from 'ilib/lib/ilib';`
But ES6 import cannot be used at runtime. It's not only just webpack/babel issue.  Browser support for the import/export section of ES6 is extremely rare. Tools like webpack and chrome convert from ES6 into ES5 and bundle the dependency tree into a large js file. For example:

// file: a.js
import {message} from './b.js';
console.log(message);

// file: b.js
const message = 'Hello';
export {message};

Would be converted by webpack/barbel to something along the idea of:

// file: app.js

(function(modules) {
... // webpack runtime code with "__webpack_require__" acting internally as the require function
return __webpack_require__(0);
})([
(function(module, exports, __webpack_require__) {
var _b = __webpack_require__(1);
var message = b;
console.log(b);
}),
(function(module, exports, __webpack_require__) {
var message = "Hello"
module.exports.message = message;
})

]);

With ES6 imports not supported natively and the 'ilib/lib/ilib/' path nonexistent for Enact apps, the current format will be not doing anything other than throwing errors in the browser.

With iLib being encapsulated within the app.js bundle, untouchable from external scripts, we'll need to modify the locmaker output format so that it's just storing the data to a global object of some sort (such as window.iLibData or something attached to a window). Then when the app runs normally, the app (or Enact itself) can set that global jsonJS object to the internal copy of ilib.data inside the app scope.

### DIFF
--- a/tools/locmaker/locmaker.js
+++ b/tools/locmaker/locmaker.js
@@ -130,7 +130,8 @@ var classList = [
 	"DateRngFmt",
 	"LocaleInfo",
 	"ResBundle",
-	"Locale"
+	"Locale",
+	"ScriptInfo"
 ];
 
 var localeList = ["ar-AE","ar-BH","ar-DJ","ar-DZ","ar-EG","ar-IQ","ar-JO","ar-KW","ar-LB","ar-LY",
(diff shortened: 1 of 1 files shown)


gouniLee:
Closing this PR. Because it's not an appropriate change to be applied.

Issue #62, opened 2017-11-15
RataDie calculation


gouniLee:
*node v0.10*
`new Date(2019, 2, 31)`
`Sat Mar 30 2019 23:00:00 GMT-0100 (AZOT)`

*node v6.11*
`new Date(2019, 2, 31)`
`2019-03-31T01:00:00.000Z`

*latest version of chrome (62.0.3202.62)*
`new Date(2019, 2, 31)`
`Sun Mar 31 2019 01:00:00 GMT+0000 (AZOST)`

 result of 
     ***getTimeExtended()***
https://github.com/iLib-js/iLib/blob/development/js/lib/RataDie.js#L282
*node v0.10*  - **1553994000000**:Sun Mar 31 2019 01:00:00
> var value = new GregorianDate({year:2019, month:3, day:31, timezone:'local'})
undefined
> var date = DateFactory(value)
undefined
> date.getTimeExtended()
1553994000000

*node v6.11*  - **1553990400000** : Sat Mar 30 2019 23:00:00
> var value = new GregorianDate({year:2019, month:3, day:31, timezone:'local'})
undefined
> var date = DateFactory(value)
undefined
> date.getTimeExtended()
1553990400000

*latest version of chrome (62.0.3202.62)*  - **1553990400000** :Sat Mar 30 2019 23:00:00
> const myValue = new GregorianDate({year:2019, month:3, day:31, timezone:'local'})
> const myDate = DateFactory(myValue)
> console.log(myDate.getTimeExtended())
> 1553990400000

Seonghyup:
I guess that iLib `RataDie` is based on old javascript engine.

Seonghyup:
We need to fix this logic(https://github.com/iLib-js/iLib/blob/development/js/lib/GregorianDate.js#L160). 

    // if dst is defined, the intrinsic Date object has no way of specifying which version of a time you mean
    // in the overlap time at the end of DST. Do you mean the daylight 1:30am or the standard 1:30am? In this
    // case, use the ilib calculations below, which can distinguish between the two properly
    var d = new Date(this.year, this.month-1, this.day, this.hour, this.minute, this.second, this.millisecond);
    this.offset = -d.getTimezoneOffset() / 1440;

Maybe there was a change of the way to judge time(daylight/standard).

gouniLee:
Closing this issue because it's fixed.

PR #65, opened 2017-11-20
Fix a bug that returning wrong time when starting DST

Seonghyup:
Fix a bug that returning wrong time when starting DST
- Align with the way to applying DST in javascript native engine when it is started on starting boundary time. 

more details: https://github.com/iLib-js/iLib/issues/62


Seonghyup:
Because of the issue below, value of `getTimeExtended()` is not correct when using `DateFactory` not `GregorianDate`.
https://github.com/enyojs/iLib/issues/19

PR #72, opened 2017-12-06
Implement AlphabeticIndex feature

gouniLee:
This class handles alphabetic indexes which are collated sequences of
 * buckets into which elements are placed sorted appropriately to the given
 * language. An example would be an index of person names in a contact
 * list, organized by the first letter of the family name.

```
 Example in English:
 Buckets: A B C D E F G H I J K L M N O P Q R S T U V W X Y Z #<p>
```
```
 A
    Adams
    Albers
    Alvarez
 B
    Baker
    Banerjee
    Brunshteyn
 ...
```
**API Lists:**
>getLocale(), addElement(), addLabels(), clear(), getAllBuckets(), getBucket(), getIndexStyle(), getBucketCount, getBucketLabels(), getAllBucketLabels(), getCollator(), getOverflowLabel(), getElementCount(), getUnderflowLabel(), setOverflowLabel(), setUnderflowLabel(), getUnderflowLabel()


js/lib/AlphabeticIndex.js
```diff
@@ -221,128 +353,303 @@ AlphabeticIndex.prototype.addLabels = function(labels, start) {
...
+
+	if (!collationValue && this.inherit) {
+		for (var i = 0; i < this.inherit.length; i++) {
+			if (this.inherit[i] === 'this') {
+				this.inherit[i] = this.style;
+    		}
+
+    		if (this.collation[this.inherit[i]].map[firstChar]) {
+    			collationValue = this.collation[this.inherit[i]].map[firstChar];
+    			this.inheritStyle = this.inherit[i];
```
ehoogerbeets: This is changing an instance variable with a temporary value. For safety, temporary values should be passed as parameters to other functions/methods so that they do not change if some other code calls a different method. Because JS is single-threaded, it is hard to get it wrong, but when you store temporary values in an instance variable and multiple methods can change it, then the value of the variable depends on which method was called last.

Much better would be to create a super-set map during initialization so that this work only has to be done once. For example, let's say a collation for hangul inherits latin. (inherit: ["this", "latin"]) Hangul has primary values that go from 27 to 238 (211 values). Latin has primary values that go from 0 to 69 (70 values). We build a new map with primary values that go from 0 to 280 and store that as an instance variable that doesn't change for the life of the instance. (Values 0 to 210 are Hangul and 211 to 280 are Latin) Then, when we want to look up a value, we check the one single map...

js/lib/AlphabeticIndex.js
```diff
@@ -221,128 +353,303 @@ AlphabeticIndex.prototype.addLabels = function(labels, start) {
...
- * 
+ *
  * @returns {Array.<String>} the array of bucket labels
  * for this index in collation order
  */
 AlphabeticIndex.prototype.getBucketLabels = function() {
+	var buckets = this.getAllBuckets();
+	var label = new Array();
+	var i;
+
```
ehoogerbeets: We can make this more efficient like this:

```
return this.getAllBuckets().map(function(bucket) {
  return bucket.label;
});
```

js/lib/AlphabeticIndex.js
```diff
@@ -221,128 +353,303 @@ AlphabeticIndex.prototype.addLabels = function(labels, start) {
...
+	if (this.allBucketLabels) {
+		return this.allBucketLabels;
+	}
+
+	this.allBucketLabels = new Array();	
+	this.allBucketLabels.push(this.underflowLabel);
+
+	for (i=0; i < this.indexUnits.length; i++) {
+		this.allBucketLabels.push(this.indexUnits[i]);
```
ehoogerbeets: concat is faster, as it it implemented natively:

```
this.allBucketLabels = [this.underflowLabel].concat(this.indexUnits, this.overflowLabel);
```

js/lib/AlphabeticIndex.js
```diff
@@ -60,159 +60,291 @@ var Collator = require("./Collator.js");
...
+			this.index[i].elements.push(element);
+			this.index[i].elements.sort();
+			newItem = false;
+			break;
+		}
+	}
+
+	if (newItem) {
+		itemSet.elements.push(element);
+		this.index[this.index.length] = itemSet;
```
ehoogerbeets: Should use "push" here. 

```
this.index.push(itemSet);
```

js/lib/AlphabeticIndex.js
```diff
@@ -60,159 +60,291 @@ var Collator = require("./Collator.js");
...
+	var newItem = true;
+	var itemSet = {};
+
+	itemSet.label = label;
+	itemSet.elements = [];
+
+	for (i = 0; i < this.index.length; i++) {
+		if (this.index[i].label === label) {
+			this.index[i].elements.push(element);
+			this.index[i].elements.sort();
```
ehoogerbeets: This needs to use the collation to sort, otherwise you are sorting in Unicode. Also, you do not need to sort every time an element is added. The elements should be sorted only when they are returned to the caller in getAllBuckets().

js/lib/AlphabeticIndex.js
```diff
@@ -221,128 +353,303 @@ AlphabeticIndex.prototype.addLabels = function(labels, start) {
...
 /**
  * Return the collator used to sort elements in this
  * index.
- * 
+ *
  * @return {Collator} the ilib Collator instance used
  * in this index
  */
 AlphabeticIndex.prototype.getCollator = function() {
+	return this.collation;
```
ehoogerbeets: This is not of type Collator. This instance variable is initialized to the collation.json in the constructor. This should return a real Collator instance.

js/lib/AlphabeticIndex.js
```diff
@@ -60,159 +60,291 @@ var Collator = require("./Collator.js");
...
+		object: "AlphabeticIndex",
+		locale: this.locale,
 		name: "collation.json",
-		sync: sync,
-		loadParams: loadParams, 
+		sync: this.sync,
+		loadParams: this.loadParams, 
+
 		callback: ilib.bind(this, function (collation) {
 			if (!collation) {
```
ehoogerbeets: We need to be creating a Collator instance here instead of just loading the collation.json data. We will use the collator in a few places in this code anyways. The map can be extracted from the collator instance anyways.

js/lib/AlphabeticIndex.js
```diff
@@ -60,159 +60,291 @@ var Collator = require("./Collator.js");
...
-							}
-		    			}));
-	    			} else {
-						if (options && typeof(options.onLoad) === 'function') {
-							options.onLoad(this);
-						}
-	    			}
-	    		})
-			});
+
```
ehoogerbeets: We use CType below, but we don't initialize it here in the constructor to allow it to load its json files.

js/lib/AlphabeticIndex.js
```diff
@@ -60,159 +60,291 @@ var Collator = require("./Collator.js");
...
+/**
+ * @private
+ */
+AlphabeticIndex.prototype._normalizeHangul = function(value) {
+	if (!value) {
+		return "";
+	}
+	var source = new NormString(value);
+	var normString = source.normalize("nfkd");
```
ehoogerbeets: Can we use NormString._decomposeHangul() directly instead of normalizing the whole string?

js/lib/AlphabeticIndex.js
```diff
@@ -221,128 +353,303 @@ AlphabeticIndex.prototype.addLabels = function(labels, start) {
...
+			overflowIndex = i
+			break;
+		}
+	}
+
+	if (overflowIndex > -1) {
+		temp = this.index.splice(overflowIndex,1)[0];
+		this.index[this.index.length] = temp;
+	}
+
```
ehoogerbeets: Need to sort the elements in each bucket here using a collator instance.

js/data/locale/collation.json
```diff
@@ -17,6 +17,60 @@
...
+			"^": [36],
+			"_": [37],
+			"`": [38],
+			"{": [39],
+			"|": [41],
+			"}": [42],
+			"~": [43],
+			"": [44],
+			"[": [59],
+			"\\": [60],
```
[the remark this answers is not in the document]
gouniLee: All of the data in a symbol is from another collation data. I extract symbol data from others collation data. so first "\\": [34], is from ascii and "\\": [60], is from latin'
The reason I made "symbol" part, for using mixed script cases. but It's not using it. It's better to remove.

js/lib/AlphabeticIndex.js
gouniLee: It's not normalizing the first string not the whole string. so I think it's ok. no?

js/lib/AlphabeticIndex.js
gouniLee: I got your considerations. 
In order to create one single map for dealing mixed script case, There's one more issue which is 'flowBoundaries' Every collation map has different values...  and What if they have the same value on both maps? such as double Backslash(\\). it is in 'ascii' and latin'

js/lib/AlphabeticIndex.js
```diff
@@ -221,128 +386,322 @@ AlphabeticIndex.prototype.addLabels = function(labels, start) {
...
+    
+    if (!element) {
+        return undefined;
+    }
+
+    source = new NormString(element);
+    firstChar = source.str[0];
+
+    if (CType.withinRange(firstChar, "hangul")) {
+        firstChar = this._normalizeHangul(firstChar);
```
ehoogerbeets: Is it possible that firstChar is already a Jamo? Does that matter?

js/lib/AlphabeticIndex.js
```diff
@@ -60,159 +62,322 @@ var Collator = require("./Collator.js");
...
+            isDigit._init(this.sync, this.loadParams, ilib.bind(this, function(){
+                new Collator ({
+                    locale: this.locale,
+                    useNative: false,
+                    sensivitiy: "primary",
+                    usage: "sort",
+                    sync: this.sync,
+                    loadParam : this.loadParams,
+                    onLoad: ilib.bind(this, function(collation){
+                        this.collationObj = collation;
```
ehoogerbeets: Should we call NormString.init here to load the files needed? If so, NormString.init should be fixed so that it only initializes the first time it is called. The second time, it should not do all the work over again.

js/lib/AlphabeticIndex.js
```diff
@@ -221,128 +386,322 @@ AlphabeticIndex.prototype.addLabels = function(labels, start) {
...
+    var charNum, firstBoundaryChar, endBoundaryChar, firstCharNum, endCharNum;
+    var source;
+    var validMapNum = -1;
+    
+    if (!element) {
+        return undefined;
+    }
+
+    source = new NormString(element);
+    firstChar = source.str[0];
```
ehoogerbeets: What you have on these two lines is equivalent to

```
firstChar = element[0];
```

... because the code doesn't do anything with the string once it is contained within the NormString instance. Specifically, it does not perform any normalization on the string "element" before looking at the first character. Why create a NormString instance on a string if you are not going to normalize that string?

Instead, this should call 

```
source = new NormString(element);
source.normalize("nfc"); 
var it = source.charIterator();
firstChar = it.next();
```

To put the string into a standard form first and take the first character of that normalized string.

js/lib/AlphabeticIndex.js
ehoogerbeets: Also, you should pass in the NormString instance from above on line 529 so that _normalizeHangul does not have to make a new instance.

js/lib/AlphabeticIndex.js
gouniLee: I just want to make sure it means changing the implementation of NormString.init() same as CType._load(), Did I get it exactly?

js/lib/AlphabeticIndex.js
gouniLee: Yes, It's possible but not very common case.

js/lib/AlphabeticIndex.js
ehoogerbeets: Yes, exactly!

gouniLee: oh, I see. so I'll re-update codes.

PR #73, opened 2017-12-12
Add new json files for Country class

Seonghyup:
- Fix bug below of Country class by adding new json files:

    When merging locale data, the key is unique of the object(except country name data) and there is no 
    problem. But in case of `Country` class, one more country names are able to be merged for a code 
    because the country name acts the key of object. For example, there is `"남한" : "KR"` in `zh` locale 
    and there is `"대한민국" : "KR"` in `zh-Hant-TW` locale. After merging this data, `Country` class returns 
    `남한` for `KR` code in `zh-Hant-TW` locale because `남한` and `대한민국` act like key in merging 
    process and iLib has country name data for `zh-Hant-TW` locale like this: 

        ...
        "남한" : "KR",
        "대한민국" :  "KR",
        ...


js/lib/Country.js
```diff
@@ -81,22 +81,22 @@ var Country = function (options) {
 
 	this.locale = this.locale || new Locale();
 	localeinfo = new LocaleInfo(this.locale);
```
ehoogerbeets: Need to load this asynchronously in case Country itself is loaded asynchronously:

```
new LocaleInfo(this.locale, {
  loadParams: loadParams,
  sync: sync,
  onLoad: function(li) {
    this.locale = li;
etc.
```

js/lib/Country.js
```diff
@@ -108,45 +108,45 @@ var Country = function (options) {
...
  * @static
  * @return {Object} an object of country code that this copy of ilib knows about.
  */
 Country.getAvailableCode = function() {
 	var ret = [],
-		country,
+		code,
 		countries = new ResBundle({
-			name: "ctrynames"
+			name: "ctryreverse"
```
ehoogerbeets: Not sure why we have to load in the same data as a ResBundle here when the data is already loaded on line 92.

js/lib/Country.js
```diff
@@ -108,45 +108,45 @@ var Country = function (options) {
...
 		}).getResObj();
-	
-	for (country in countries) {
-		if (country && countries[country]) {
-			ret.push(country);
+
+	for (code in countries) {
+		if (code && countries[code]) {
+			ret.push(countries[code]);
```
ehoogerbeets: There should be a private method that calculates the reverse (it sets this.countryToCode) which should be called here. Then, this function would be just like the other one above:

```
if (!this.countryToCode) {
  this._calculateCountryToCode();
}
return Object.keys(this.countryToCode);
```

js/lib/Country.js
```diff
@@ -157,32 +157,32 @@ Country.prototype = {
...
 
-				this.codeToCountry = {};
+				this.countryToCode = {};
 				this.locinfo = li;
 
-				for (ctry in temp) {
-					if (ctry && temp[ctry]) {
-						this.codeToCountry[temp[ctry]] = ctry;
+				for (code in temp) {
+					if (code && temp[code]) {
```
ehoogerbeets: Call this._calculateCountryToCode() here too.

js/lib/Country.js
```diff
@@ -108,45 +108,45 @@ var Country = function (options) {
...
+			name: "ctryreverse"
 		}).getResObj();
-	
-	for (country in countries) {
-		if (country && countries[country]) {
-			ret.push(countries[country]);
+
+	for (code in countries) {
+		if (code && countries[code]) {
+			ret.push(code);
```
ehoogerbeets: Much more efficient:

```
return Object.keys(this.codeToCountry);
```

The reason is that Object.keys is implemented natively.

tools/cldr/gencountrynames.js
```diff
@@ -32,7 +31,7 @@ var mergeAndPrune = common.mergeAndPrune;
 var makeDirs = common.makeDirs;
 
 function usage() {
-    console.log("Usage: gencountrynames [-h] CLDR_json_dir locale_data_dir\n" +
+    console.log("Usage: gencountrymnames [-h] CLDR_json_dir locale_data_dir\n" +
```
ehoogerbeets: What doe the "m" stand for?

Seonghyup: Oh, It is mistake. I will modify it.

js/lib/Country.js
Seonghyup: I agree.

js/lib/Country.js
Seonghyup: `getAvailableCode` is static function, so it need to load data. 
Is it better to save data in `ilib.data`?

js/lib/Country.js
Seonghyup: I have a question.. `getAvailableCountry` is static method. Is it possible to call private method in static method(when `Country` instance does not exist)?

js/lib/Country.js
```diff
@@ -80,9 +80,15 @@ var Country = function (options) {
...
-	if (localeinfo.getRegionName() === undefined) {
+	new LocaleInfo(this.locale, {
+		sync: sync,
+		loadParams: loadParams,
+		onLoad: ilib.bind(this, function (li) {
+			this.locinfo = li;
+		})
+	});
+
+	if (this.locinfo.getRegionName() === undefined) {
```
ehoogerbeets: The rest of this constructor has to appear inside of the onLoad callback on line 86 or else this.locinfo will be undefined on line 91 during async operation. The onLoad function will be called when the new LocaleInfo is created, which may occur after this constructor is finished running.

PR #75, opened 2017-12-19
Return a date component string for UI builders 

ehoogerbeets:
Added a method getDateComponentOrder() to the date formatter class to aid UI widget developers. This returns a three-letter string like "ymd" which gives the order of the components needed when making a date selector widget. The developer can then use this to put the UI components for year, month, and date in the correct order.

The developer can combine this call with calls like DateFmt.getMonthsOfYear() to get the names of all the months, and methods of the Calendar class to get the number of months, and the lengths of each month in a particular year.

### DIFF
--- a/js/lib/DateFmt.js
+++ b/js/lib/DateFmt.js
@@ -699,6 +699,13 @@ DateFmt.prototype = {
 					this.template = this._getFormat(this.formats.time[this.clock], this.timeComponents, this.length);
 					break;
 			}
+			
+			// calculate what order the components appear in for this locale
+			this.componentOrder = this._getFormat(this.formats.date, "dmy", "l").
+			    replace(/[^dMy]/g, "").
+			    replace(/y+/, "y").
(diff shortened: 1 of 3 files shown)


js/lib/DateFmt.js
```diff
@@ -902,6 +909,30 @@ DateFmt.prototype = {
...
+	 * of the three letters have at least one locale that uses that ordering, though some
+	 * combinations are far more likely than others. The ones that are not used by any 
+	 * locales are "dym" and "myd", though new locales are still being added to 
+	 * CLDR frequently, and possible orderings cannot be predicted. Your code should 
+	 * support all 6 possibilities, just in case.
+	 * 
+	 * @return {String} a string giving the date component order
+	 */
+	getDateComponentOrder: function() {
+	    return this.componentOrder;
```
Seonghyup: How about returning `Array`? I think it is easier than `String` to handle.

PR #76, opened 2018-01-11
Add the Locale.getLangSpec() method

ehoogerbeets:
Makes it slightly more useful. Now you can get a string representing the written language for the locale.


gouniLee:
I have a question. We have LocaleMatcher feature. It matches full Locale Spec with likelocales.json. For example, "ca" --> "ca-Latn-ES" Does it slightly conflict this feature?

ehoogerbeets:
This is a convenience function for things outside of ilib. For example, Amazon Elastic Search, SOLR, and Lucene all require the language or language+script as the locale for indexing and searching documents. The reason is that the region doesn't matter when matching two documents in the same language (and for most languages, the script doesn't either.) This is just a method for returning that language + optional script locale instead of calculating it again and again elsewhere.

gouniLee:
I see. Thank you for the additional explanation!

PR #77, opened 2018-01-11
Update the locale matching functionality

ehoogerbeets:


gouniLee:
When I run test case  by executing  root/nodeunit/testSuite.html.  Many of them failed. Could you please to check?

```
Cannot read property 'zxx' of undefined undefined (actual) undefined undefined (expected)
TypeError: Cannot read property 'zxx' of undefined
    at LocaleMatcher._getLikelyLocale (http://127.0.0.1:9090/output/js/ilib-ut.js:26152:43)
```

gouniLee:
I have few questions.

1. Definition 'macrolanguage'.
I can understand how 'macrolanguage' is used. When doing a match two locales, It checks macroLanguage values. and It two locale have same macroLanguageReverse value, similarity score[0] 90. Regarding the definition of 'macrolanguage'.  can I say it is kind of default language fallback value?

2. macrolanguages.json,  likelyLocalesAdditioal.json and mutualIntelligibility.json
I realize those three json files are manually added. Not coming from cldr. Could you leave a link or additional information that you reference? Especially I wonder how to extract mutualiIntelligibility score. i.e. cs-sk: 85,   It doesn't cover all locale cases. Don't  we need to update that case?

gouniLee:
Thank you for the additional explanation!
Regarding likelyLocalesAdditioal.json, In CLDR, has data such as, `und-AD` in, Can we think it as same as `AD`?

ehoogerbeets:
I remember now why I put the likely locales additional file. The reason is that many of the CLDR things are wrong. For example, CLDR has "und-HK" mapping to "en-Latn-HK" whereas I have it mapping to "zh-Hant-HK". English is widely spoken in Hong Kong, but the most common language is Chinese. Many of my additional ones are overlaps, and I could remove them in the future.

gouniLee:
Yes, I agree.  Next time we can check the latest version of cldr. and it can be removed if some are duplicated. By the way, I don't know which version are you referring to. I'm looking at version 30.and It's written "und-HK"  is mapping to "zh-Hant-HK"

PR #84, opened 2018-02-23
Fix asynchronous calls to work properly

ehoogerbeets:
Changes:

- make sure everything in ilib uses the callbacks instead of calling other ilib classes synchronously
- added support for real asynchronous mode on nodejs using fs.readFile() instead of fs.readFileSync(). This is implemented in the new AsyncNodeLoader class.
- changes to the ant build and the test frameworks to support this async mode of testing
- added unit tests for most classes to make sure they load their locale data properly
- added tests for browser-based async testing (though you will not be able to run these until the webpack work is done.)
- for many classes, if they throw an exception in sync mode, make sure to call the callback with an undefined argument to indicate the error when in async mode
- fixed many whitespacing issues. Slowly we will convert all tabs to 4 spaces only.


ehoogerbeets:
The builds have all passed now. This is ready to be reviewed.

gouniLee:
Wow, huge change! 
I saw you created new AsncNodeLoader for Node. which uses `fs.readFile` instead `fs.readFileSync`  I wonder how about the browser? Does it use same WebLoader for async feature with use xhr as `sync:false` options? In order to check this part, I treid `testSuiteAsync.html` but none of them not running.  `test/[feature]/nodeunit/testSuiteAsync.html` Could you please to check again?

gouniLee:
'test/[feature]/nodeunit/testSuiteAsync.html' will work `webpack loader`. I've heard it would work in `webpack branch`.

PR #90, opened 2018-03-11
Update to IANA timezone data 2018c

ehoogerbeets:


ehoogerbeets:
Also fixed up the gentzdata.js tool to use the cldr-data module, and fixed a bug in the zoneinfo compiler.

PR #100, opened 2018-04-20
Fix unit formatting

ehoogerbeets:
This PR is ready for review now. Read the changes to docs/tutorial/units.md first to get an idea of what the whole change is before looking at the code.


gouniLee:
It looks like big change. and It would be nicer to give more accurate formatting for users. :) 
For unformatting, iLib gets data from CLDR. and CLDR defined 3 lengths (narrow, short and long) but the iLib doesn't use narrow length for unit formatting. Is there any you intention? Because other feature uses narrow data from CLDR.

gouniLee:
Oh! Have you run test cases on QT/QML? It fails some cases in `unit` and `util` category.

ehoogerbeets:
All Qt tests are fixed now. Please try again.

gouniLee:
Yes, regarding my question, it is not tightly related this PR. I just found that while looking UniFmt code.
I checked All of cased are Passed on QT/QML. Thanks!

PR #105, opened 2018-05-10
Add support for significant digits parameter to the number formatter

ehoogerbeets:
Truncates the number at the N most significant digits. This is similar to maxFractionDigits, but significantDigits applies to the whole number, not just the fractional part.

Now you can write:

```
var nf = new NumFmt({significantDigits: 3});
console.log(nf.format(12345678));
// output: 12,300,000
console.log(nf.format(0.0012345678));
// output: 0.00123
```

If there is a conflict, then whichever creates a shorter number will win.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -1,13 +1,19 @@
 Release Notes for Version 13.0
 =============================
-Build 003
+
+Build 004
 -------
 Published as version 13.3.0
 
 New Features:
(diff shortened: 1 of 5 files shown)


ehoogerbeets:
Yes, I love food features. They are tasty.

gouniLee:
oh! I was supposed to write 'good` though. but you know what I meant already. ;-)

PR #118, opened 2018-07-03
Update DST time calculation logic (inDaylightTime())

gouniLee:
If you set the time with 15:30 and timezone "local", it does depend on the current time zone, yet. 
it will automatically calculate the offset to UT in getTimeZoneExtended().

If return value in getTimeZoneExtended() change a day, and the day is DST start/end date affect DST calculation logic. In order to fix this case,  I added logic to compare the offset value of the date minus the offset time.

fyi. [PLAT-61832] https://jira2.lgsvl.com/browse/PLAT-61832


gouniLee:
Closing. It's related https://github.com/iLib-js/iLib/pull/119

PR #120, opened 2018-07-07
Add ability to get the default timezone from the Intl object

ehoogerbeets:
If the Intl object is available on this platform, ilib.getTimeZone() will attempt to get it from there. If it can't, then it will fall back to its previous behaviour of checking environment variables, and if there are none, then fall back to the hard coded defaults.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -6,10 +6,12 @@ Build 004
 Published as version 13.3.0
 
 New Features:
-* LocaleReference Documentation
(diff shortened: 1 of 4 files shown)


gouniLee:
As far as I know, `Intl` have a little bit different local fallback mechanism. with `lookup` and `bestfit` option. and it might have different result per browser. Do you think its ok that we can directly use `Intl.DateTimeFormat().resolvedOptions` ?

gouniLee:
This feature is no locale dependency because it uses the utc time and timezone of the currently running machine.

PR #121, opened 2018-07-09
Add option.sync is true when it is undefined in DateFmt

gouniLee:
Fix legacy platform enyo 2.6+ bug:
Previously, that options were `true` for initial object setting. but it's undefined now.
If it's `undefined`. It set as `false` eventually in https://github.com/iLib-js/iLib/blob/development/js/lib/Utils.js#L341
It occurs error during load `dateformats.json`  in legacy platform enyo 2.6+

### DIFF
--- a/js/lib/DateFmt.js
+++ b/js/lib/DateFmt.js
@@ -621,7 +621,7 @@ DateFmt.prototype = {
      * Finish initializing the formatter object
      */
     _init: function(options) {
-        if (!options.sync) {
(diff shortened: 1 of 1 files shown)


gouniLee:
Oooooooohhhh sooory. I accidentally I clicked the merge button! I was supposed to merge https://github.com/enyojs/iLib/pull/126. NOT this PR...;( If you don't agree this change. Please let me know...

ehoogerbeets:
This one is fine, yes.

PR #124, opened 2018-09-04
Add getFormatInfo() to AddressFmt

ehoogerbeets:
This method will return the information that user interfaces need to present a locale-sensitive input form for addresses.


js/lib/AddressFmt.js
```diff
@@ -208,4 +209,209 @@ AddressFmt.prototype.format = function (address) {
...
+ *     "label": "Street Address"
+ *   }],
+ *   [{
+ *     "component": "locality",
+ *     "label": "City"
+ *   },{
+ *     "component": "region",
+ *     "label": "State",
+ *     "constraint": {
+ *       "AL": "Alabama",
```
odeniso: JS objects don't preserve an order here, right? might be better to implement this as an array of label/value tuples

`[["AL", "Alabama"], ["AK", "Alaska"]] ...`

ehoogerbeets: Almost all engines in modern browsers preserve order (Firefox, Chrome, Edge) and also in recent versions of nodejs, Qt, and Rhino, but the spec doesn't mandate that you have to preserve in for..in loops, even in ES2015. So there is a de-facto standard but not an explicit standard. There are only a few older and lesser-used browsers that still do not preserve order. The rest of ilib already relies on order preservation, so this is a bigger issue than just this object. 

BUT, your comment here does bring up a good point. The order is not correct per locale! It is currently sorted according to the original English string. I think I will fix that in a subsequent change though to keep this one small.

gouniLee:
Approved. I've confirmed every test cases are passed between node, browser, and QT. 5.4. We should add more test cases.

I'm looking at file regionnames.json, (it's not a related question of this PR)
the key name is different per Country. some of them are number (even I couldn't guess the rule) and some of them are Alphabetic. Where are this data come from? I couldn't find tool code either.

gouniLee:
oh. yes, I realize a file name combine with filePrefix +"names.json"  
I checked the `cldr-json` data from  https://github.com/unicode-cldr/cldr-json and I couldn't get data name like '서울특별시' neither I wonder why that data doesn't provide json type. 
For Korea, those kinds of number match with region name doesn't familiar. so I was curious.

PR #128, opened 2018-09-13
Moji 708: support sorted lists in AddressFmt.getFormatInfo()

ehoogerbeets:
When getFormatInfo returns a list of states/provinces/prefectures/etc. or it returns the list of countries, it now returns them sorted properly for the locale.

In order to do that, we needed to change the API a bit. Instead of returning an object with sorted properties and relying on the engine to preserve that order, now we return an array of small objects, each having a code and name property. The reason is that objects with sorted properties are fine on most engines, but only if those properties do not contain any numbers. If there are numbers, those numbered properties get returned first, out of order. The state/province/etc. region codes for some countries contain a mix of numbers and letter, creating unsorted lists, and requiring this solution.

Since this API is not released yet, we still have the opportunity to change the return value like this without breaking anyone.


ehoogerbeets:
A lot of files have changed, but all the code is in the last 5 files in the PR. Ignore all of the data changes, and scroll all the way down to the bottom.

gouniLee:
I didn't know `every` method. It's good to know  :) Browser tests aren't working. but I'm sure it would be ok when a webpack branch is merged

PR #136, opened 2018-09-28
Make QT/QML build and test work properly.

gouniLee:
QT/QML build and test work properly as fixing qt/buld.xml file
Qt build is working in ilib whole build process, and ant test.all is working.

But, I've found some tests failed in duration while qt/qml test


qt/build.xml
```diff
@@ -354,8 +333,7 @@ limitations under the License.
 		
 	<!-- Should not run the tests from an automated build, because qmlscene doesn't exit at the end. They should
 	     only be run by hand. -->
-	<target name="test" description="Run all tests and build all reports" />
-
+	<target name="test" depends="" description="Run all tests and build all reports" />
```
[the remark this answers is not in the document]
gouniLee: I guess accidentally copied it. It means nothing.  I'll remove it

qt/build.xml
```diff
@@ -351,11 +322,5 @@ limitations under the License.
             </exec>
         </sequential>
     </target>
-		
-	<!-- Should not run the tests from an automated build, because qmlscene doesn't exit at the end. They should
-	     only be run by hand. -->
 	<target name="test" description="Run all tests and build all reports" />
-
-	<target name="doc" />
```
ehoogerbeets: The top-level build.xml has a "doc" target which iterates through all the subdirectories and calls "doc" there. If this target does not exist, that target will fail at the top level. Since there are no docs here, it is implemented as nothing, but it at least has to be here so the overall build doesn't fail.

gouniLee: oh.. I didn't know that. you're right. So I'll put that code back.

PR #144, opened 2018-10-18
Creating a new top-level entry point for ilib

ehoogerbeets:
Before: we had to load in the top-level ilib first to cause it to install the NodeLoader. If you didn't do that first, then the other ilib classes do not know how to load json files.

```
// installs NodeLoader to load json files
var ilib = require("ilib");  

// now that ilib is required above, the code below can now load json data using NodeLoader
var AddressFmt = require("ilib/lib/AddressFmt");
```

There are a number of problems with the above.

1. You have to load ilib first, which may be useless in the rest of the code if it doesn't use the ilib object directly. Plus, eslint complains about the unused variable.
2. It always installs NodeLoader -- what if you were were on Rhino or Qt or more importantly, within Webpack? That's the wrong loader.
3. If you are on one of the other platforms, you have to explicitly use require("ilib/lib/ilib-qt") for Qt for example. What if you didn't know where the code was going to be run before it is run? Or what if the same code was going to be run on multiple platforms?
4. If you forgot to load ilib first, the other classes will silently work but use the US English defaults, so you don't even know it isn't working correctly until later when some sort of class fails or you don't get the results you expect.

After:

```
var AddressFmt = require("ilib/lib/AddressFmt"); // automatically uses the correct loader to load json
```

AddresFmt depends on the new index.js top-level entry point. This new entry point checks the current environment (node, browser, webpack, qt, etc) and loads the appropriate loader for that env.

That way, you can use "npm install ilib" to get ilib onto your local disk and then use it in node with NodeLoader. Then, when you webpack that same code, it will automatically use the
WebpackLoader instead.

### DIFF
--- a/js/build.xml
+++ b/js/build.xml
@@ -342,8 +342,10 @@ limitations under the License.
             <fileset dir="${build.lib}">
                 <include name="ilib-node*"/>
                 <include name="ilib-qt*"/>
-                <include name="ilib-getdata.js"/>
                 <include name="ilib-unpack.js"/>
+            </fileset>
+            <fileset dir="${build.base}">
+                <include name="index.js"/>
             </fileset>
 			<mapper type="glob" from="*.js" to="*.js"/>
 		</apply>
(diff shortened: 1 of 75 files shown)


ehoogerbeets:
Needed to merge the loader and plugin changes first and publish them on npm before the CI build of ilib will succeed.

PR #147, opened 2018-10-24
Work around problem where uglify was removing the !defineLocaleData comment

ehoogerbeets:
Without the !defineLocaleData comment, the plugin and loader do not have
any target where they could put their output and therefore there is no
dependency on the locale data, which therefore never gets included into
the webpacked code.

The problem was that webpack thought the code block with only the comment
in it was empty, so the dead-code optimizer removed that entire block,
including the comment. The solution was to put the comment in another
file and require that file in that block so that the dead code optimizer
does not think that block is empty.


qt/FileReader/.qmake.stash
```diff
@@ -16,3 +16,30 @@ QMAKE_MAC_SDK.macx-clang.macosx.QMAKE_RANLIB = \
 QMAKE_MAC_SDK.macx-clang.macosx.QMAKE_LINK = /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++
 QMAKE_MAC_SDK.macx-clang.macosx.QMAKE_LINK_SHLIB = /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++
 QMAKE_MAC_SDK.macosx.platform_name = macosx
+QMAKE_MAC_SDK.macosx.--show-sdk-path = /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.13.sdk
```
jeeyi: Do you need this file to be in git? Looks like a tmp file to me.

PR #153, opened 2018-10-26
[Not to merge] Enable qt test on travis. Just for testing 

gouniLee:
Enable iLib QT test (except collation feature) on Travis.  It is to see if it still does not work.


gouniLee:
still doesn't work. closing.

PR #166, opened 2018-12-10
Update the Taiwan area code of Phone Number according to Wikipedia

gouniLee:
### Checklist

According to Wikipedia, I've updated area code of Phone Number information in zh-Hant-TW. I got a report the `039` is not area code. and I re-searched a information and updated. and Additionally, I've fixed 3 digits `iddprefix` number format by removing `-` at the end. 

### Resolution
Re-generated `json` file related `zh-Hant-TW` by using tool code. I've fixed all of the related test cases.

### Additional Considerations
[//]: # (How should the change be tested?)
[//]: # (Are there any outstanding questions?)
[//]: # (Were any side-effects caused by the change?)

### Links
https://en.wikipedia.org/wiki/Telephone_numbers_in_Taiwan
https://de.wikipedia.org/wiki/Telefonvorwahl_(Taiwan)


docs/ReleaseNotes.md
```diff
@@ -1,6 +1,15 @@
 Release Notes for Version 14
 ============================
 
+Build 005
+-------
+Published as version 14.1.3
```
ehoogerbeets: If you put this in the release notes, then please update the build.properties and the package.json to 14.1.3 as well.

PR #175, opened 2019-03-06
Move index.js to lib/index.js for enyojs

ehoogerbeets:
### Checklist

### Resolution

### Additional Considerations
[//]: # (How should the change be tested?)
[//]: # (Are there any outstanding questions?)
[//]: # (Were any side-effects caused by the change?)

### Links

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -13,4 +13,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-version=14.1.5
+version=14.1.4
--- a/build.properties
+++ b/build.properties
@@ -1,4 +1,4 @@
-# Copyright © 2012-2018, JEDLSoft
+# Copyright © 2012-2019, JEDLSoft
 # 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -13,4 +13,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-version=14.1.4
+version=14.1.5
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -1,6 +1,14 @@
 Release Notes for Version 14
 ============================
 
+Build 006
+-------
+Published as version 14.1.4
+
+Bug Fixes:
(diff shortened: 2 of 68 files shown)


gouniLee:
With this change, `Uncaught RangeError: Maximum call stack size exceeded error, ` problem is fixed. 
But there's another issue. For Enyo or Enact perspective, The loader provided by iLib side, a need not be loaded automatically. They have their own. but `index.js` The iLib judge currently running platform is `webos` and try to load  `./ilib-node.js` as a default which doesn't fit for enyo or enact.
Do you need to provide not loading any type of loader? or add additional logic to figure out specific platform?

ehoogerbeets:
Maybe enyo should provide a new npm module like "enyo-ilib-main" or something like that. All it would be is two files. One is the WebosLoader.js file, and the other is index.js file that does this:

```javascript
var WebosLoader = require("./WebosLoader.js");
var ilib = require("ilib/lib/ilib.js");

ilib.setLoaderCallback(WebosLoader(ilib));

ilib._dyncode = true; // indicate that we are using dynamically loaded code
ilib._dyndata = true; // and data

module.exports = ilib;
```

Then when the various ilib classes require index.js, it will test that ilib._dyncode and ilib._dyndata are booleans already and will not load anything.

Does that work?

Older enyo things could continue to use ilib 13.x

gouniLee:
As a general approach, Shouldn't we suggested a way not to use a type of Loader that iLib provided?
For example, If developers use iLib on node environment, but they've written their own NodeLoader. they want to use their loader instead of iLib's NodeLoader. I think we need to set a guideline for that first. 
or could we add more check option  `if statement` in [index.js](https://github.com/iLib-js/iLib/blob/development/js/index.js#L22)? ( I can't think of any proper option now though.)  If not, let people create their npm module that you suggested. That could be a possible option.
After that, I think I can decide how to solve the problem issue regarding enyo apps.

Modifying any enyo apps are risky on webOS. If I have to ask to change any part related enyo, I would choose just using current version instead of updating with the latest one..

gouniLee:
I've decided not updating iLib to Enyo for webOS. I will provide iLib legacy version for enyo framework. So This PR doesn't need to be merged.

ehoogerbeets:
Abandoning...

PR #180, opened 2019-05-13
Re-introduce merged data caching for QT

ehoogerbeets:
- other platforms can use it as well, but only QT does it by default

### Checklist

### Resolution

### Links

### DIFF
--- a/js/lib/Utils.js
+++ b/js/lib/Utils.js
@@ -384,8 +384,6 @@ Utils.loadData = function(params) {
         throw "Utils.loadData called without a callback. It must have a callback to work.";
     }
 
-    if (ilib._getPlatform() === "qt") ilib._cacheMerged = true;
-
     if (params.name) {
         name = params.name;
     }
--- a/js/lib/Utils.js
+++ b/js/lib/Utils.js
@@ -116,40 +116,44 @@ var Utils = {};
 Utils.getSublocales = function(locale) {
     var ret = ["root"];
     var loc = typeof(locale) === "string" ? new Locale(locale) : locale;
+    var lang = loc.getLanguage();
+    var region = loc.getRegion();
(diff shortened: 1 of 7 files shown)


gouniLee:
if the `ilib._cacheMerged` value is set properly,  performance is improved with your change.  but still worse than before. When the fist creating the object, it takes much time than before. 
I noticed the line below is changed too,
https://github.com/iLib-js/iLib/blob/development/js/lib/DateFmt.js#L435
Previously,  it was 
`DateFactory._dynLoadDate(this.calName); `
it looks like you've changed to webpack support. Could we improve this part too?

js/lib/Utils.js
```diff
@@ -384,6 +384,8 @@ Utils.loadData = function(params) {
         throw "Utils.loadData called without a callback. It must have a callback to work.";
     }
 
+    if (ilib._getPlatform() === "qt") ilib._cacheMerged = true;
+
```
ehoogerbeets: All direct manipulation of the ilib name space should be contained within the ilib.js file as much as possible.

js/lib/ilib.js
```diff
@@ -794,6 +794,6 @@ ilib.isDynData = function() {
  * merging is too slow, so the already-merged data is cached as well after the first time it
  * is requested. In this way, we sacrifice the memory footprint for the sake of speed.
  */
-if (ilib._getPlatform() !== 'qt') ilib._cacheMerged = false;
+ilib._cacheMerged = false;
```
ehoogerbeets: Should be `ilib._cacheMerged = (ilib._getPlatform() === 'qt');`

gouniLee: yeah..but If I add this line `ilib.js`. It makes lots of test case failures.

PR #198, opened 2019-07-10
Test what a codecove is..

gouniLee:
No merge. Only for testing...
https://docs.codecov.io/docs/flags

### Checklist

### Resolution

### Links


gouniLee:
Closing the PR.  I've looked through what a codecov is. I expected to easily integrate into iLib. but In order to do that, I need to extract coverage data from nodeunit test framework. especially as `lcov` style. `nodeunit` has an option to get that report, but it doesn't work.. and there're some issues and probably need more time. It would be nicer if we have, but not worth to spend much time to get it.

PR #199, opened 2019-07-18
Update numplan.json files based on  metadata version 1.7.20

gouniLee:
### Checklist

I've generated all the `numplan.json` files through `gennumplan.js` with libphonenuber library's **metadata version 1.7.20.**
I can't get a `dialingPlan` information from that. so It has `closed` as default. this value will need to be checked.
I think Following country lists are having verified data especially  regarding 'dialingPlan'.
```["KR", "US", "GB", "ES", "MX", "AR", "CO", "BR", "CA", "FR", "IT", "DE", "RU", "JP", "CN", "TW", "NL", "AU", "IN", "BE", "LU", "IE", "NZ", "HK"];```
To avoid breaking test cases, I haven't generated data for the following countries. and just adding country code if it doesn't have it

Updated country list for `_normPhoneReg` in `PhoneLocale.js` following metadata.json from libphonenumber-js library. and Modified some countries based on Wikipedia information.

### Resolution

### Links

### CODE REVIEW

File: js/data/locale/und/CA/numplan.json [hunk 1]
```diff
@@ -0,0 +1,21 @@
+{
```
ehoogerbeets: There should be no numplan.json for all of the North American Numbering Plan (NANP) countries other than the US. There is code that maps these countries to the US numplan. This includes Canada, Bermuda, Virgin Islands, etc. Also, some countries normalize to France and some to Italy. See https://github.com/iLib-js/iLib/blob/development/js/lib/PhoneLocale.js#L180 for all the mappings.

js/data/locale/und/GF/numplan.json
```diff
@@ -2,5 +2,20 @@
     "region": "GF",
```
ehoogerbeets: Same here... normalizes to France

js/data/locale/und/VA/numplan.json
```diff
@@ -1,8 +1,20 @@
 {
```
ehoogerbeets: Should not be overridden

js/data/locale/und/SM/numplan.json
```diff
@@ -2,5 +2,19 @@
     "region": "SM",
```
ehoogerbeets: Normalizes to Italy

gouniLee:
I see. Countries in PhonLocael.js doesn't need to have `numplan.json` 
I want to check if these lists need to be updated or not. If countries have the same country calling codes, could we say they have the same numbering system? 
Please see this info: https://github.com/iLib-js/iLib/blob/development/tools/phone/metadata.json#L15083 
The data look similar but different. I'm not sure if I'm ok to following this countries info or just leave iLib version and remove JSON files

js/data/locale/und/CN/numplan.json
```diff
@@ -1,9 +1,8 @@
 {
     "region": "CN",
-    "countryCode": "86",
```
ehoogerbeets: What happened to the country code?

js/data/locale/und/BE/numplan.json
```diff
@@ -1,9 +1,8 @@
 {
     "region": "BE",
-    "countryCode": "32",
```
ehoogerbeets: Country code is missing?

js/data/locale/und/DE/numplan.json
```diff
@@ -1,10 +1,9 @@
 {
     "region": "DE",
-    "countryCode": "49",
```
ehoogerbeets: In all of these ones that are modified, the country code is missing.

gouniLee: Oh yes. That part was I manually updated. Because the tool code did not generate information for the country, and the existing version of files did not have a country code before... I'll update again

PR #204, opened 2019-07-26
Update phoneloc.json and wrote a script file to automatically generate it

gouniLee:
### Checklist

I've written a script file to automatically generate it. Regarding an mcc*Moible country code) number, I've hardcoded information from  https://en.wikipedia.org/wiki/Mobile_country_code and use `metadata.json` from `libphonnumber-js` library
There're some test failures. The information from Wikipedia, There are duplicate numbers.505 It for Australia and  Norfolk Island. so I commented out for Norfolk Island at this moment. 

### Resolution

### Links


js/test/phone/testnormalize.js
```diff
@@ -961,7 +961,7 @@ module.exports.normalize = {
         test.done();
     },
     
-    testAssistedDialNormalizeBogusNumberForVerizonInvalidCountryCode: function(test) {
+    testAssistedDialNormalizeBogusNumberForVerizonvalidCountryCode: function(test) {
```
ehoogerbeets: This test is supposed to test against an invalid country code. Did you meant to change the name of the function?

gouniLee: Yes, you are right. it shouldn't be removed for invalid country code testing. I reverted it

PR #216, opened 2019-10-28
Add in the missing methods to IString

ehoogerbeets:
These methods have been added to the String prototype in the years since this class was first coded. Here we put in methods to delegate to the string class. The unit tests only attempt to run if the string class supports these. Older versions of node and browsers of course do not have these new methods.

### Checklist

### Resolution

### Links

### DIFF
--- a/js/lib/IString.js
+++ b/js/lib/IString.js
@@ -957,6 +957,15 @@ IString.prototype = {
     },
 
     /**
+     * Same as String.matchAll()
+     * @param {string} regexp the regular expression to match
+     * @return {iterator} an iterator of the matches
+     */
+    matchAll: function(regexp) {
+        return this.str.matchAll(regexp);
+    },
+
+    /**
      * Same as String.replace()
      * @param {string} searchValue a regular expression to search for
      * @param {string} newValue the string to replace the matches with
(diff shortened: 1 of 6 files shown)


gouniLee:
Tests are failed on QT 5.12
* QT 5.12 WebEngine : Chromium 69 (https://wiki.qt.io/New_Features_in_Qt_5.12)
* `matchAll` full support : Chrome 73 : (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll )

```
  [exec] Using Qt version 5.12.4 in /home/goun/Qt_5.12/5.12.4/gcc_64/lib
...
  [exec] qml: [runner.js] runTests: running suite /root/testSuiteFiles.js
     [exec] qml: [TestEnvironment.qml] (path) : /root/testSuiteFiles.js (moduleName): root
     [exec] qml: This version of the JS engine does not support String.matchAll()
     [exec] qml: [testStringDelegateEndsWithTrue] assertion failed: Expected that actual == [[true]] , but got [[false]] instead.
     [exec] qml: This version of the JS engine does not support String.trimEnd()
     [exec] qml: This version of the JS engine does not support String.trimRight()
     [exec] qml: This version of the JS engine does not support String.trimStart()
     [exec] qml: This version of the JS engine does not support String.trimLeft()
     [exec] qml: This version of the JS engine does not support String.matchAll()
     [exec] qml: [testStringDelegateEndsWithTrue] assertion failed: Expected that actual == [[true]] , but got [[false]] instead.
     [exec] qml: This version of the JS engine does not support String.trimEnd()
     [exec] qml: This version of the JS engine does not support String.trimRight()
     [exec] qml: This version of the JS engine does not support String.trimStart()
     [exec] qml: This version of the JS engine does not support String.trimLeft()
     [exec] qml: This version of the JS engine does not support String.matchAll()
     [exec] qml: [testStringDelegateEndsWithTrue] assertion failed: Expected that actual == [[true]] , but got [[false]] instead.
     [exec] qml: This version of the JS engine does not support String.trimEnd()
     [exec] qml: This version of the JS engine does not support String.trimRight()
     [exec] qml: This version of the JS engine does not support String.trimStart()
     [exec] qml: This version of the JS engine does not support String.trimLeft()
     [exec] qml: This version of the JS engine does not support String.matchAll()
     [exec] qml: [testStringDelegateEndsWithTrue] assertion failed: Expected that actual == [[true]] , but got [[false]] instead.
     [exec] qml: This version of the JS engine does not support String.trimEnd()
     [exec] qml: This version of the JS engine does not support String.trimRight()
     [exec] qml: This version of the JS engine does not support String.trimStart()
     [exec] qml: This version of the JS engine does not support String.trimLeft()
     [exec] qml: This version of the JS engine does not support String.matchAll()
     [exec] qml: [testStringDelegateEndsWithTrue] assertion failed: Expected that actual == [[true]] , but got [[false]] instead.
     [exec] qml: This version of the JS engine does not support String.trimEnd()
     [exec] qml: This version of the JS engine does not support String.trimRight()
     [exec] qml: This version of the JS engine does not support String.trimStart()
     [exec] qml: This version of the JS engine does not support String.trimLeft()
     [exec] qml: This version of the JS engine does not support String.matchAll()
     [exec] qml: [testStringDelegateEndsWithTrue] assertion failed: Expected that actual == [[true]] , but got [[false]] instead.
     [exec] qml: This version of the JS engine does not support String.trimEnd()
     [exec] qml: This version of the JS engine does not support String.trimRight()
     [exec] qml: This version of the JS engine does not support String.trimStart()
     [exec] qml: This version of the JS engine does not support String.trimLeft()
     [exec] qml: This version of the JS engine does not support String.matchAll()
     [exec] qml: [testStringDelegateEndsWithTrue] assertion failed: Expected that actual == [[true]] , but got [[false]] instead.
     [exec] qml: This version of the JS engine does not support String.trimEnd()
     [exec] qml: This version of the JS engine does not support String.trimRight()
     [exec] qml: This version of the JS engine does not support String.trimStart()
     [exec] qml: This version of the JS engine does not support String.trimLeft()
     [exec] qml: This version of the JS engine does not support String.matchAll()
     [exec] qml: [testStringDelegateEndsWithTrue] assertion failed: Expected that actual == [[true]] , but got [[false]] instead.
     [exec] qml: This version of the JS engine does not support String.trimEnd()
     [exec] qml: This version of the JS engine does not support String.trimRight()
     [exec] qml: This version of the JS engine does not support String.trimStart()
     [exec] qml: This version of the JS engine does not support String.trimLeft()

```

gouniLee:
I did a debugging with a test failures regarding `assertion failed: Expected that actual == [[true]] , but got [[false]] instead.` case.
For example, `str.endsWith("def")`
https://github.com/iLib-js/iLib/blob/istringUpdate/js/test/root/teststrings.js#L1129
It expects return `true` but it's `false`
If I try `str.str.endsWith("def")`  then it returns `true`

The line here https://github.com/iLib-js/iLib/blob/istringUpdate/js/lib/IString.js#L52
I've seen a message `Do not use "String" as a constructor"
https://doc.qt.io/qtcreator/creator-checking-code-syntax.html
http://linterrors.com/js/do-not-use-a-as-a-constructor

ehoogerbeets:
@gouniLee try now. I changed the `new String()` to just `String()`

gouniLee:
@ehoogerbeets  I've updated code to use `toString()` It's hard to debug QT inside at this time to figure out the reason to shows the wrong result. With that change, all test is passed. Following methods are not supported on QT5.12  (`matchAll(), trimEnd(), trimRight(), trimStart(), trimLeft()` If you agree this, It's ok to be merged.

js/lib/IString.js
```diff
@@ -1068,6 +1068,12 @@ IString.prototype = {
      * the end of the string, and false otherwise
      */
     endsWith: function(searchString, length) {
+        /* (note)length is optional. If it is omitted the default value is the length of string.
+        *  But If length is omitted, it returns false on QT. (tested on QT 5.12)
+        */
+        if (typeof length === "undefined") {
+            length = this.str.length;
+        }
```
ehoogerbeets: Nice find!

gouniLee:
I tested a bit more,
when `length` is omitted it returns the wrong value.
For example:
```
var text = "abcdef"
text.startsWith("abc") // true
text.startsWith("abc", undefined) // true

text.endsWith("def") // true
text.endsWith("def", undefined) // false
```

ehoogerbeets:
Should make those 4 into unit tests then.

ehoogerbeets:
Looks good to me! Okay to merge this?

PR #218, opened 2019-12-06
Implement a "familiar" name format

ehoogerbeets:
This is as familiar as you can get and yet not be rude to a stranger for each locale.

### Checklist

### Resolution

### Links


js/data/locale/ko/name.json
```diff
@@ -1,7 +1,8 @@
 {
 	"components": {
 		"formal_short": "pgf",
-		"formal_long": "pgf"
+		"formal_long": "pgf",
+        "familiar": "pgf"
```
laughnan: `[nit]` tab formatting here

.gitignore
```diff
@@ -23,3 +23,4 @@ qt/FileReader/release/
 qt/output/
 
 /node_modules/
+/dist/
```
laughnan: did you mean to include this?

ehoogerbeets: yes, that is the destination for the build script when you build the "distribution"

js/test/name/testnamefmt.js
```diff
@@ -693,7 +730,25 @@ module.exports.testnamefmt = {
         test.equal(fmt.format(name), "닥터 박은성");
         test.done();
     },
-    
+
+    testNameFmtKOFormalLong: function(test) {
```
ehoogerbeets: Name conflict with the previous function

js/test/name/testnamefmt.js
```diff
@@ -745,9 +746,25 @@ module.exports.testnamefmt = {
...
+            locale: "ko-KR"
+        });
+        var fmt = new NameFmt({
+            style: "familiar",
+            locale: "ko-KR"
+        });
+
+        test.equal(fmt.format(name), "박은성 님");
+        test.done();
+    },*/
```
gouniLee: @ehoogerbeets What do you think about the set default value even developer don't put suffix? (This testcase doesn't activate at this point)

gouniLee:
@ehoogerbeets Regarding the Korean case, I've updated the component to include `suffix` It's pretty important to put 'suffix` especially not to be rude. So I want to get your opinion to give default `suffix` value.

js/test/name/testnamefmt.js
ehoogerbeets: We can do that in a separate PR

PR #223, opened 2020-02-24
Update script relate data to CLDR 36 and UCD 13.0.0

gouniLee:
### Checklist

`scriptToRange.json` `rangeToScript.json`, and `scripts.jf` files are updated to CLDR 36, Unicode 13.0.0  by running `genscritps.js`, `genlangscripts.js`.
`scripts.json` hasn't changed since the last updated version.

### Links
https://unicode.org/iso15924/iso15924-text.html
http://unicode.org/Public/cldr/36/ --> common/properties/scriptMetadata.txt


js/data/locale/bh/scripts.jf
```diff
@@ -1,7 +0,0 @@
-{
```
ehoogerbeets: I think Behari is still written in Devanagari, right? Why are these files deleted?

gouniLee:
Regarding deleted files, I think those were not in CLDR (`cldr-data/supplementa/languageData.json`) and It wasn't properly deleted last cldr34 update. It should've been deleted last time work. The hard part is that CLDR data isn't easy to figure out the change if it's not a major change.

gouniLee:
FYI.
Deleted languages are associated with data in [supplementtalMetadata.xml](https://github.com/unicode-org/cldr/blob/master/common/supplemental/supplementalMetadata.xml) 
i.e)
```
<languageAlias type="kpv" replacement="kv" reason="macrolanguage"/> <!-- Komi-Zyrian ⇒ Komi -->
```
The data associated with them use the replacement value. So kpv data is found under ‘kv’.

PR #226, opened 2020-02-25
Update numfmt.jf and currency locale data to CLDR 36 

gouniLee:
### Checklist

* [x] Requires a major version change.

Update `currency.json`, `currency.jf` and `numfmt.jf` locale data to CLDR 36 by running `gennumfmt.js` and `gencurrencies.js`. Fixed all test case failures. 

### Resolution

### Links


js/data/locale/currency.jf
```diff
@@ -1 +0,0 @@
-{"currency": "USD"}
```
ehoogerbeets: The root currency should be US Dollars if no others are specified

js/data/locale/currency.json
```diff
@@ -32,7 +32,7 @@
     "KES": {
         "name": "Kenyan Shilling",
         "decimals": 2,
-        "sign": "Sh"
+        "sign": "KES"
```
ehoogerbeets: The sign should contain the common use symbols used in the country for that currency, not the ISO code. There is another option on the number formatter to use the ISO code instead.

js/data/locale/currency.json
```diff
@@ -42,32 +42,32 @@
     "CHF": {
         "name": "Swiss Franc",
         "decimals": 2,
-        "sign": "Fr"
+        "sign": "CHF"
```
ehoogerbeets: The symbols are in the "symbol-alt-narrow" property of the cldr json files.

js/data/locale/currency.jf
gouniLee: I'll rollback the file. I'll update the extractor tool code so that it can be automatically generated from the tool code.

js/data/locale/currency.json
gouniLee: I'll update the extractor code then regenerate files.

js/test/number/testnumfmt.js
```diff
@@ -4073,7 +4073,7 @@ module.exports.testnumfmt = {
     
         test.ok(fmt);
     
-        test.equal(fmt.format(100110.57), "ƒ 100.110,57");
```
ehoogerbeets: Most of these changes are changing from double quote to single quote, which does not make any semantic difference. I would not make these sort of unnecessary changes in a PR that is mixed with actual code changes from CLDR because it hides the real changes. Things like fixing whitespace or changing quotes or other such "clean up" type of changes should go into separate PRs.

gouniLee: Yeah, I think I'd better closing this PR, then re-create PR to see changes clearly and I'll separate clean up PR if it's necessary.

PR #227, opened 2020-02-26
Update plurals.json locale data to CLDR 36 and Fix the tool code to work properly

gouniLee:
### Checklist

Update `plurals.json` locale data to CLDR 36 by running genplural.js
Regarding extractor code, I modified the code to use `cldr-data` module. and Changed tab to spaces.

### Resolution

### Links


gouniLee:
> These changes look fine, but they are missing unit tests for at least the "big" locales like "pt"

I added some test cases.

PR #235, opened 2020-03-09
Update langname and regionname file to CLDR 36 

gouniLee:
### Checklist

* Updated `langname.jf` and `regionname.jf` file to CLDR 36.
* Updated `genlangreg.js.js` tool code file to include `XX` region value automatically.
* Added newly added language to `js/build.xml` for testing
* Added testcase regarding newly added or modified data.
* Regarding removed langname.jf files, I think that language identifiers are from `ISO 639-3` which are not in CLDR. I haven't made a decision if iLib needs to keep the data inside or not

### Resolution

### Links


js/data/locale/afa/langname.jf
```diff
@@ -1,4 +0,0 @@
-{
-    "generated": true,
```
ehoogerbeets: Almost all of these deleted languages are actually collective language groups. That is, a name for a group of similar languages. It is okay to delete these.

js/data/locale/und/MK/regionname.jf
```diff
@@ -1,4 +1,4 @@
 {
     "generated": true,
-    "region.name": "Macedonia"
+    "region.name": "North Macedonia"
```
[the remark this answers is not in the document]
gouniLee: I'll check it

js/data/locale/afa/langname.jf
gouniLee: Thank you for the confirmation!

tools/cldr/genlangreg.js
```diff
@@ -28,32 +28,32 @@ var Locale = common.Locale;
...
 
 function usage() {
-    console.log("Usage: genlangreg [-h] [ toDir ]\n" +
+    console.log("Usage: genlangreg [-h] [ locale_data_dir ]\n" +
         "Generate the langname.jf and regionname.jf files for each locale.\n" +
         "-h or --help\n" +
         "  this help\n" +
-        "toDir\n" +
-    "  directory where generated files will be located\n");
+        "localeDirName\n" +
```
ehoogerbeets: that should be "locale_data_dir" so that it matches what is in the usage above

PR #240, opened 2020-03-17
Update list.json files to CLDR 36

gouniLee:
### Checklist

* Updated `list.json` file to CLDR 36 and fixed test failures and added some test cases.
* Ragarding `build.xml` file modification, I've added `en-MP` locale to list for testing.

### Resolution

### Links


js/test/strings-ext/testlistfmt_en.js
```diff
@@ -50,7 +50,7 @@ module.exports.testlistfmt_en = {
     
         test.expect(2);
         test.ok(fmt !== null);
-        test.equal(fmt.format(["one", "two"]), "one and two");
```
ehoogerbeets: It looks like CLDR changed the format for the "short" length to be a tight, small list. I don't think that is appropriate for the default length for this class. I think we should change the default length to be "medium" or "long" instead. If we do that, then many of these unit tests do not need to change.

gouniLee: I'm going to change as `medium`.  nonetheless, both `medium' and `long` cause 200+ fail cases.

js/test/strings-ext/testlistfmt_sv.js
```diff
@@ -108,7 +108,7 @@ module.exports.testlistfmt_bs = {
```
ehoogerbeets: the wrapper should be `module.exports.testlistfmt_sv` instead of `module.exports.testlistfmt_bs`

gouniLee: oh yeah, it should. I update it.

js/test/strings-ext/testlistfmt.js
```diff
@@ -2803,7 +2837,7 @@ module.exports.testlistfmt = {
 
         test.expect(2);
         test.ok(fmt !== null);
-        test.equal(fmt.format(["bir", "два", "три"]), "bir два три");
```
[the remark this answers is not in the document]
gouniLee: wow! I haven't even noticed them.I've corrected all cases that you indicated.

js/test/strings-ext/testlistfmt_pa.js
```diff
@@ -145,7 +143,7 @@ module.exports.testlistfmt_pa = {
     
         test.expect(2);
         test.ok(fmt !== null);
-        test.equal(fmt.format(["واحد", "اثنان"]), "واحد, اثنان");
+        test.equal(fmt.format(["واحد", "اثنان"]), "واحد ਅਤੇ اثنان");
```
ehoogerbeets: This is not right. The list items are in Arabic script, but the translation of "and" (ਅਤੇ) is written in the Devanagari script. I think the pa language normally uses Devanagari, but the the pa-Arab-PK locale uses Arabic. In this case, the file js/data/locale/pa/Arab/list.json is not correct.

gouniLee: The Extract tool code should generate `pa/Arab/list.json` file. but It wasn't.  It works with cldr 34 and the data haven't changed much in cldr 36. It's weird. I need to look at the tool code more.

gouniLee:
Currently, the form of `list.json` can have two styles.
One with 4 length style and another is just one. Because of that, Merging and Loading data form occur unexpected results. So Final pa-Arab's list.json's standard form is all. (Please refer the example below. ) So I think we need to unify to avoid these cases. I haven't re-generated list.json files yet. @ehoogerbeets What do you think?
```
{
     "standard": {
        "2": "...",
        "end": "...."
        "short": {
            "2": "...",
            "end": ""
        },
        "medium": {
        },
...
```

js/test/strings-ext/testlistfmt_es.js
```diff
@@ -679,7 +679,7 @@ module.exports.testlistfmt_es = {
     
         test.expect(2);
         test.ok(fmt !== null);
-        test.equal(fmt.format(["uno", "dos"]), "uno dos");
+        test.equal(fmt.format(["uno", "dos"]), "uno y dos");
```
ehoogerbeets: Update the copyright year

js/test/strings-ext/testlistfmt_fr.js
```diff
@@ -108,7 +108,7 @@ module.exports.testlistfmt_fr = {
     
         test.expect(2);
         test.ok(fmt !== null);
-        test.equal(fmt.format(["un", "deux"]), "un deux");
```
ehoogerbeets: Update the copyright year

js/test/strings-ext/testlistfmt_he_IL.js
```diff
@@ -96,7 +96,7 @@ module.exports.testlistfmt_he_IL = {
     
         test.expect(2);
         test.ok(fmt !== null);
```
ehoogerbeets: Update the copyright year

ehoogerbeets:
@gouniLee the merge code needs to be updated then to handle two cases:

```javascript
parent: "string",

child: {
    "short": "string1",
    "medium": "string2",
    "long": "string3",
    "full": "string4"
}
```

and

```javascript
parent: {
    "short": "string1",
    "medium": "string2",
    "long": "string3",
    "full": "string4"
},

child: "string",
```
In both cases, the child is more specific, so its information should be the one that is merged, even if one or more of the strings in the object is the same as the single string. The merge code in datefmts.js should be updated. This may affect (and fix!) other data types as well, so we need to check it.

gouniLee:
> @gouniLee the merge code needs to be updated then to handle two cases:
> 
> ```js
> parent: "string",
> 
> child: {
>     "short": "string1",
>     "medium": "string2",
>     "long": "string3",
>     "full": "string4"
> }
> ```
> 
> and
> 
> ```js
> parent: {
>     "short": "string1",
>     "medium": "string2",
>     "long": "string3",
>     "full": "string4"
> },
> 
> child: "string",
> ```
> 
> In both cases, the child is more specific, so its information should be the one that is merged, even if one or more of the strings in the object is the same as the single string. The merge code in datefmts.js should be updated. This may affect (and fix!) other data types as well, so we need to check it.

yeah.. I agree with your saying. This work will be done on another branch.

ehoogerbeets:
Let's go ahead and change the cldr tool to generate the full data structure for every locale, then do the merge, and then introduce a new step to optimize the fully merged data. The optimize step looks at the values of the short/medium/long/full and if they are all the same, then replace them with a single string. That way, we can keep the structure small, but now the merge works properly. Make sense? We probably need to do that for any of the other cldr tools if ilib supports these short-cut data structures...

gouniLee:
After cldr optimize a process, I think the problem will still occur. doesn't it? In different scripts under one language. For example, `pa/list.json, pa/Arab/list.json`
If all full,long,medium,short data are the same in pa/list.json so it can be optimized as one.
but pa/Arab/list.json has the difference between full and short. so we can't make it as one.

gouniLee:
I've re-generated all of `list.json` file and all have 4 length styles now. (The tool code has updated already) so It solved merging data cases and passed all of the test cases.

ehoogerbeets:
Oh, one thing I forgot to add before I approved this... we need something in the release notes that everything was updated to CLDR 36. I guess the release notes update can be a separate PR that covers all the things that were updated.

gouniLee:
The release note has been updated to the previous PR. It's written as `Updated all locale data to CLDR 36 and Unicode Character Database 13.0` for next release version `14.6.0` Do I need to write down the details feature name?

PR #248, opened 2020-05-28
Put base option back on ResBundle

Seonghyup:
Restore base option on ResBundle.
But I guess that base option is supported by loader before, we need to add base option to other feature (like DateFmt, DurationFmt..) or modify loader to support base option.
I think it is good idea to allow the use of base option in ResBundle only.

### Checklist

### Resolution

### Links

### DIFF
--- a/js/lib/ResBundle.js
+++ b/js/lib/ResBundle.js
@@ -250,8 +250,8 @@ var ResBundle = function (options) {
 
         if (typeof(options.loadParams) !== 'undefined') {
             this.loadParams = options.loadParams;
-            if (typeof (options.loadParams.root) !== 'undefined') {
-                this.path = options.loadParams.root;
+            if (typeof (options.loadParams.base) !== 'undefined') {
+                this.path = options.loadParams.base;
             }
         }
         if (typeof(options.missing) !== 'undefined') {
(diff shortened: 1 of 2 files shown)


js/lib/ResBundle.js
```diff
@@ -250,8 +250,8 @@ var ResBundle = function (options) {
 
         if (typeof(options.loadParams) !== 'undefined') {
             this.loadParams = options.loadParams;
-            if (typeof (options.loadParams.root) !== 'undefined') {
-                this.path = options.loadParams.root;
```
ehoogerbeets, line 254: For backwards compatibility, we need to support both root and base in the loadParams.

Seonghyup: Oh, I had misunderstood(I thought that this PR will be merged to v15). I've fixed it.
I have question about base option policy. Do you mean that the use of base option is allowed in ResBundle only after v14? Before v14, all classes using loader can use base option.

ehoogerbeets: It will go into v14. The problem was that we removed it from v13->v14 without deprecating it first. It needs to be deprecated for at least one major version (v14) before it can be removed. (Sometimes 2 major versions if major users like LG still need it.)

PR #251, opened 2020-08-20
Add test cases to verify eu-ES(Basque-Spain)

gouniLee:
### Checklist

* Added test cases to verify eu-ES (Basque-Spain)
* timezone: Europe/Madrid
* Added test case feature: LocaleInfo,Script, String(plural) LocaleMatch,  DateFmt, DurFmt, DateRngFmt, List, Unit, firstDayofWeek, clock, calendar,Month/Weekday Translation, NumFmt,

### Resolution

### Links


js/test/date/testcalendar.js
```diff
@@ -23,14 +23,14 @@ if (typeof(LocaleInfo) === "undefined") {
 
 module.exports.testcalendar = {
     testDefaultCalendar_GregorianCaseAll: function(test) {
-        test.expect(193);
+        test.expect(195);
         var localeList = ["ar-AE","ar-BH","ar-DJ","ar-DZ","ar-EG","ar-IQ","ar-JO","ar-KW","ar-LB","ar-LY","ar-MA","ar-MR",
-        "ar-OM","ar-QA","ar-SA","ar-SD","ar-SY","ar-TN","ar-YE","as-IN","az-Latn-AZ","bg-BG","bn-IN","bs-Latn-BA,bs-Latn-ME",
+        "ar-OM","ar-QA","ar-SA","ar-SD","ar-SY","ar-TN","ar-YE","as-IN","az-Latn-AZ","be-BY","bg-BG","bn-IN","bs-Latn-BA,bs-Latn-ME",
```
ehoogerbeets, line 28: Are all these locale test branches independent of each other? If so, be-BY should not appear in this list in this branch

js/test/date/testdatefmt_eu_ES.js
```diff
@@ -0,0 +1,1788 @@
...
+            locale: "eu-ES",
+            year: 2011,
+            month: 9,
+            day: 29,
+            hour: 13,
+            minute: 45,
+            second: 0,
+            millisecond: 0
+        });
+        test.equal(fmt.format(date), "2011(e)ko ira. 29(a)");
```
ehoogerbeets, line 81: From perusing various basque web sites, it seems that the (e) and the (a) are optional and are used when formatting the correct gender/case of the number. Sort like when you write "word(s)" in English to indicate a possible plural. That means they are optional parts that depend on the number that is being formatted. I am not sure this date format is acceptable to Basque people, or if it looks really awkward. Please have this verified by any Basque LG members you can find.

gouniLee: Regarding `eu-ES` locale, it's not officially required to support at this locale now, and it's just preparing for reference platform perspective. so It's hard to find(or contact) Basque people. The data is directly from CLDR without modifying it. I think It's better to ask to CLDR.

js/test/date/testcalendar.js
gouniLee: Actually, I wanted to separate PRs per local to see the change clearly. but it modifies the same files at the same time and can't expect when it's merged. Probably it's not a good idea to make a PR per locale.. I need to make test cases for 8 locales. probably making one PR would be better.

js/test/date/testdatefmt_eu_ES.js
ehoogerbeets: Different PRs are better. It is easier to see the differences, and merging the one file is not so difficult.

gouniLee: ok. I'll do that way.

PR #260, opened 2020-09-16
Fix DateRng formatting to present standAlone case

gouniLee:
### Checklist

According to DateRange Formatting spec, https://github.com/iLib-js/iLib/blob/development/docs/DateRangeFormatSpec.md
When the Month and Year are displayed as like the case 'c20`, it should present the standAlone value of the month if it exists.
I've made fixed to support that and fixed unit tests.

### Resolution

### Links
http://cldr.unicode.org/translation/date-time-1/date-time-patterns#TOC-When-to-use-Standalone-vs.-Formatting
http://cldr.unicode.org/translation/date-time-1/date-time

### DIFF
--- a/js/lib/DateRngFmt.js
+++ b/js/lib/DateRngFmt.js
@@ -374,6 +374,13 @@ DateRngFmt.prototype = {
         monthTemplate = this.dateFmt._tokenize(this.dateFmt._getFormatInternal(formats, "m", this.length) || "MM");
         dayTemplate = this.dateFmt._tokenize(this.dateFmt._getFormatInternal(formats, "d", this.length) || "dd");
 
+
+        /*
+        * Some languages use two different forms of strings (standlone and format) depending on the context.
+        * Typically the standalone version is the nominative form of the word,
+        * and the format version is in the genitive (or related form).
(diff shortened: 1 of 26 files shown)


js/lib/DateRngFmt.js
```diff
@@ -372,6 +374,10 @@ DateRngFmt.prototype = {
         monthTemplate = this.dateFmt._tokenize(this.dateFmt._getFormatInternal(formats, "m", this.length) || "MM");
         dayTemplate = this.dateFmt._tokenize(this.dateFmt._getFormatInternal(formats, "d", this.length) || "dd");
 
+        if (isStandAlone && monthTemplate[0].length >= 3) {
```
ehoogerbeets, line 377: Need a comment here to explain why this is. I can just imagine that a few years from now, we are looking at this and thinking to ourselves, "What is this?!?!"

ehoogerbeets:
Need a release note + update version number

js/lib/DateRngFmt.js
gouniLee: yes, you're right. I added an explanation and updated version number.

ehoogerbeets (approved):

PR #261, opened 2020-09-25
Implement day periods (ie. the "B" format letter in date formats)

ehoogerbeets:
### Checklist

### Resolution

### Links


js/test/date/testMonthTranslation.js
```diff
@@ -498,18 +498,18 @@ module.exports.testmonthtranslation = {
...
-        test.equal(value[3], "Απρίλιος");
-        test.equal(value[4], "Μάιος");
-        test.equal(value[5], "Ιούνιος");
-        test.equal(value[6], "Ιούλιος");
-        test.equal(value[7], "Αύγουστος");
-        test.equal(value[8], "Σεπτέμβριος");
-        test.equal(value[9], "Οκτώβριος");
-        test.equal(value[10], "Νοέμβριος");
-        test.equal(value[11], "Δεκέμβριος");
+        test.equal(value[0], "Ιανουαρίου"); //standAlone
```
gouniLee, line 501: According to cldr 36, Modified values are wrong.
"stand-alone" ->
"wide": {
                  "1": "Ιανουάριος",
                  "2": "Φεβρουάριος",
                  "3": "Μάρτιος",
                  "4": "Απρίλιος",
                  "5": "Μάιος",
                  "6": "Ιούνιος",
                  "7": "Ιούλιος",
                  "8": "Αύγουστος",
                  "9": "Σεπτέμβριος",
                  "10": "Οκτώβριος",
                  "11": "Νοέμβριος",
                  "12": "Δεκέμβριος"
                }

js/test/date/testdatefmt_az_Latn_AZ.js
```diff
@@ -771,7 +771,7 @@ module.exports.testdatefmt_az_Latn_AZ = {
             second: 0,
             millisecond: 0
         });
-        test.equal(fmt.format(date), "Sentyabr 2011");
+        test.equal(fmt.format(date), "sentyabr 2011");
```
gouniLee, line 774: In this formatting, Sentyabr(standAlone) is correct, no?

js/test/date/testdatefmt_bs_Cyrl_BA.js
```diff
@@ -540,7 +540,7 @@ module.exports.testdatefmt_bs_Cyrl_BA = {
             second: 0,
             millisecond: 0
         });
-        test.equal(fmt.format(date), "9");
+        test.equal(fmt.format(date), "с");
```
gouniLee, line 543: If the date type is `m` and 'short` length, it should be the digit. no?

js/test/daterange/testdatefmtrange_fmttemplate.js
```diff
@@ -2795,7 +2795,7 @@ module.exports.testdaterangefmt = {
         test.equal(templatec11[0], '{sm} {sd} – {em} {ed}, {ey}');
         test.equal(templatec11[1], '{sm} {sd} – {em} {ed}, {ey}');
         test.equal(templatec11[2], '{sm} {sd} – {em} {ed}, {ey}');
-        test.equal(templatec11[3], '{sm}/{sd}/{sy} – {em}/{ed}/{ey}');
+        test.equal(templatec11[3], '{sm}/{sd} – {em}/{ed}/{ey}');
```
gouniLee, line 2798: According to spec documentation, https://github.com/iLib-js/iLib/blob/development/docs/DateRangeFormatSpec.md
cFmt11, short length daterange format is
`c11 | short | Mdy –Mdy`
Do you want to change this part?

js/test/durfmt/testdurfmt.js
```diff
@@ -4943,7 +4943,7 @@ module.exports.testdurfmt = {
             second: 1,
             millisecond: 1
         });
-        test.equal(duration.toString(), "1년 1개월 1주 1일 1시간 1분 1초 1밀리초");
+        test.equal(duration.toString(), "1년 1개월 1주 1일 1시간 1분 1초 1ms");
```
gouniLee, line 4946: Actually, I've been manually fixed this part to Korean, even though the CLDR provides the `ms`.

js/test/date/testdatefmt_az_Latn_AZ.js
ehoogerbeets: Not in az locale. If you look at the availableFormats for az, the yMMM and yMMMM use the M format letter, not the L, so it uses the regular spelling of the month name. It's only the MMM and MMMM formats that use the L format letter.

js/test/date/testMonthTranslation.js
ehoogerbeets: If you look in CLDR at the Greek date formats file, you will see that the availableFormats MMM and MMM use the regular month names instead of the stand-alone versions. If you look at the yMMM and yMMMM, that is where the stand-alone names are used. Previously, if the stand-alone month names are different than the regular month names, we used the stand-alone everywhere, which is not correct. With this change, the MMM and yMMM can be different.

In this case above, the format was MMMM, so the unit tests are now switched to the regular names instead of the stand-alone versions.

js/test/date/testdatefmt_bs_Cyrl_BA.js
ehoogerbeets: In Bosnian, the "m" date component uses the stand-alone now. So, instead of the "m" date component mapping to the "M" format, it now maps to the "L" format, which uses letters instead of digits.

js/test/daterange/testdatefmtrange_fmttemplate.js
ehoogerbeets: In the {date}{time} version of Mdy (line 56), it used `Mdy - Mdy` and in the {time}{date} version of Mdy (line 119), it used `Md - Mdy`. One or the other was wrong! I thought the one on line 56 was the one that was wrong because the year is supposed to be the same for the c11 format, so it makes sense to share it. I will update this spec to match.

js/test/durfmt/testdurfmt.js
ehoogerbeets: Then I will introduce code into the cldr tool to override this then.

js/test/date/testMonthTranslation.js
gouniLee: oh.. If you so, My previous PR, https://github.com/iLib-js/iLib/pull/260 is also wrong.
Because I forced to switch from a character the `M` to `L` when the c20 length duration formatting.

js/test/date/testdatefmt_bs_Cyrl_BA.js
gouniLee: little confused. What if a user wants to display with a number in that case?

ehoogerbeets: Then they would not use a date component specifier. The `date: "m"` means "give the format where the month name appears alone". If they wanted only the number, they would have to use the `template` property

gouniLee: then what is https://github.com/iLib-js/iLib/blob/development/js/lib/DateFmt.js#L853 line for? 
In the root dateformats.json we have `n` (I think it should be `l`) others have  'l' (ru/dateformats.json)  I thought this character is for display standalone, and  character `m'  is to display digits.  no?

ehoogerbeets:
This is ready for re-review now.

js/lib/DateFmt.js
```diff
@@ -842,14 +851,16 @@ DateFmt.prototype = {
...
     // stand-alone of w (weekday) is e
     // stand-alone of y (year) is r
     _standAlones: {
-        "l": "m",
-        "a": "d",
-        "e": "w",
-        "r": "y"
+        "m": "l",
+        "my": "mys",
+        "d": "a",
```
gouniLee, line 861: oh I see, standAlone for `my` is newly added. :)  It's pretty much clean up. Regarding the format, `a` Isn't it ok to be removed? It look like there's no format data and test case too.

js/lib/DateFmt.js
```diff
@@ -852,13 +852,11 @@ DateFmt.prototype = {
 
     // stand-alone of m (month) is l
     // stand-alone of my (month year) is mys
-    // stand-alone of d (day) is a
     // stand-alone of w (weekday) is e
     // stand-alone of y (year) is r
     _standAlones: {
         "m": "l",
         "my": "mys",
-        "d": "a",
```
gouniLee, line 854: oh.. Sorry for the confusion.;( We need to keep day's standAlone data. There are related test cases. such as Japanese, Korean.. we need to keep that info, probably  extract from cldr. `availableFormats-> d`?

gouniLee (approved):

PR #262, opened 2020-09-25
Remove IString dynamic length

pesimeao:
### Checklist

* [X] `ReleaseNotes` has added or is not needed.
* [X] This PR has passed all of the test cases on `Nodejs` and `Chrome Browser`.
* [X] This PR has passed all of the test cases on `QT/QML`.

The IString implementation breaks on webOS 3.x devices (WebKit 537.36) with the following error: `Cannot redefine property: length`.
This happens while loading the file.
image

User agent: `Mozilla/5.0 (Web0S; Linux/SmartTV) AppleWebKit/537.36 (KHTML, like Gecko) QtWebEngine/5.2.1 Chrome/38.0.2125.122 Safari/537.36 WebAppManager`

### Resolution
Since the string is a constant, there is no need to define a getter for this. The value computed in the constructor should be enough and no break changes were done.

### DIFF
--- a/js/lib/IString.js
+++ b/js/lib/IString.js
@@ -532,6 +532,7 @@ IString.prototype = {
      * console.log("String is " + str._length() + " characters long.");
      * </pre>
      * @private
+     * @deprecated
      */
     _length: function () {
         return this.str.length;
--- a/js/lib/IString.js
+++ b/js/lib/IString.js
@@ -1444,10 +1444,4 @@ IString.prototype = {
     }
 };
 
-Object.defineProperty(IString, 'length', {
-    get: function() {
-        return this._length();
-    }
-});
-
 module.exports = IString;


js/lib/IString.js
```diff
@@ -1444,10 +1444,4 @@ IString.prototype = {
     }
 };
 
-Object.defineProperty(IString, 'length', {
-    get: function() {
-        return this._length();
```
ehoogerbeets, line 1449: The _length() method on line 536 is now unused. Please remove that as well to clean up.

pesimeao: I haven't done that because that would be a break change if someone is using it (I know it is private, but still...).
What do you think if we add the `deprecated` notation?

ehoogerbeets: Okay, good plan

pesimeao: Thank you. Updated.

ehoogerbeets (approved):

PR #264, opened 2020-10-13
Updated tool code to generate `timezone.jf` file

gouniLee:
### Checklist

* Updated `gentzinfo.js` code to generate `timezone.jf` file that uses `zone.tab` from tzdata. If there are multiple timezone IDs exist in the country, it chooses the one as a reference to current ilib data.

### Resolution

### Links


tools/cldr/gentzinfo.js
```diff
@@ -218,4 +219,50 @@ fs.writeFile(filepath, JSON.stringify(countryToZones, true, 4), function (err) {
...
+    "US": "America/New_York"
+}
+
+for (var country in countryToZones) {
+    var directory = path.join(toDir, "und", country);
+    if (!fs.existsSync(directory)) {
+        mkdirs(directory);
+    };
+    var data = {
+        timezone: typeof defaultTimeZoneID[country] !== "undefined" ? defaultTimeZoneID[country]: countryToZones[country][0]
```
ehoogerbeets, line 262: Here is where we need to read in the tzdata file called "backward" and map from the LINK-NAME column to the TARGET column.

ehoogerbeets (changes requested):

ehoogerbeets (approved):
Fix the unit tests, and we should be good to go

PR #265, opened 2020-10-14
Fix "C" locale and no-region locales

ehoogerbeets:
- If the platform has the locale "C", treat it like "en-US"
- If the platform does not specify a locale, treat it like "en-US"
- If the locale uses a locale that only has the language, don't
misparse it

### Checklist

### Resolution

### Links


gouniLee (approved):
:+1::+1:

PR #273, opened 2020-11-13
Change CLDR tools to use Unicode json instead of cldr-data

ehoogerbeets:
Made sure that the output from these scripts is the same as what is in ilib now. The only difference is that gendatefmt2.js produces its properties in a different order.

### Checklist

### Resolution

### Links


gouniLee (approved):
Shouldn't we update the release notes regarding this change?

ehoogerbeets:
Well, there is no change to anything that is shipped, so release notes aren't really necessary.

PR #274, opened 2020-11-23
Added guard code for edge case in JSUtils.merge not to throw an error

gouniLee:
### Checklist

Added guard code to consider when the object is `undefined` in `JSUtil.merge`  at least not to throw an error.

### Resolution

### Links


ehoogerbeets (approved):
Need a release note?

PR #275, opened 2020-11-23
Change cldr tools to use ucd-full and iso-15924 packages instead of downloaded UCD and ISO txt files 

ehoogerbeets:
No updates to the ilib code or data were done, only the tools. (That's the next PR!) Verified that these tools produce the same output as the old tools did with UCD 12.

### Checklist

### Resolution

### Links


tools/cldr/gencoll.js
```diff
@@ -72,79 +62,44 @@ process.argv.forEach(function (val, index, array) {
...
-        console.error("Could not access file " + scriptFileName);
-        usage();
-    }
-});
+if (!fs.existsSync(toDir)) {
+    console.error("Could not access target directory " + toDir);
+    usage();
+}
 
 ducetFileName = cldrDir + "/common/uca/allkeys_DUCET.txt";
```
gouniLee, line 82: Because there is no npm module containing this file, couldn't be updated, right?

ehoogerbeets: Correct. Also, gencoll.js doesn't really work. I never finished it. Instead, I did the data files by hand. It was a pain in the you-know-what, but it was faster than trying to figure out the stupid ambiguous unicode file syntax.

gouniLee (approved):
Approved. there are no ilib code and data changes. believe the output of the tool code is the same as before.

PR #276, opened 2020-12-09
Update data to UCD 13/CLDR 37

ehoogerbeets:
### Checklist

### Resolution

### Links


tools/cldr/genunits.js
```diff
@@ -118,6 +118,34 @@ function isAsianLanguage(locale) {
     return l.getLanguage() === "ja" || l.getLanguage() === "zh" || l.getLanguage() === "th";
 }
 
+// filter out these relative units
+var filter = {
+    "10p-1": true,
```
gouniLee, line 123: what are these keys mean? Does cldr contain related data? I tried to find it, but I couldn't

ehoogerbeets: You can see these at the top of this file: https://github.com/unicode-cldr/cldr-units-full/blob/master/main/en/units.json

They are used to construct metric names. For example, you combine "meter" with the string 10p3 (=1000) which is "kilo{0}" to get "kilometer". We don't use those, so I filtered them out.

gouniLee (approved):

PR #283, opened 2021-03-19
Implement Locale.isValid() method

ehoogerbeets:
Checks to see if the language, script, and region are all valid
ISO codes and returns false if not.

### Checklist

### Resolution

### Links


js/lib/Locale.js
```diff
@@ -578,6 +578,212 @@ Locale.a1toa3langmap = {
     "zu": "zul"
 };
```
gouniLee, line 580: I guess it's from https://unicode.org/iso15924/iso15924-codes.html 
How about leaving a link to where this data came from? (In case of a list change)

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,17 @@
 Release Notes for Version 14
 ============================
 
+Build 013
+-------
+Published as version 14.8.0
+
+New Features:
+* Added Locale.isValid() method which returns true iff all of the given locale components (language,
```
gouniLee, line 9: iff?

ehoogerbeets: "iff" means "if and only if": https://en.wiktionary.org/wiki/iff

js/lib/Locale.js
ehoogerbeets: Yes, it came from there originally. I copied it from js/data/locale/scripts.json ... I wanted to make sure that the Locale code could run without loading any data, which would require async operation in some situations. Requiring async operation for `new Locale` is not something I wanted to happen.

Anyways, yes, I'll put that link in the comment.

docs/ReleaseNotes.md
gouniLee: oh, I see.  it's really new to me :)

gouniLee (approved):

PR #297, opened 2021-06-09
Fix locale parsing in ilib.getLocale()

ehoogerbeets:
* Fixed various bugs parsing the platform locales in ilib.getLocale()
    * Locales with a script code such as "zh-Hans-CN"
    * The posix "C" default locale
    * Platforms where the region code is not upper-case
    * Platforms that don't use a dash to separate the components
    * Platforms that include a dot and a charset name after the specifier
* Now includes many tests for getLocale that simulates the environment of various platforms while running under node, so we can test everything under node as well as on the real platform.

### Checklist

### Resolution

### Links

### DIFF
--- a/js/lib/ilib.js
+++ b/js/lib/ilib.js
@@ -288,7 +288,7 @@ ilib.setLocale = function (spec) {
  * @private
  */
 function parseLocale(str) {
-    if (!str || str === "C") return str;
+    if (!str) return str;
 
     // take care of the libc style locale with a dot + script at the end
     var dot = str.indexOf('.')
@@ -299,6 +299,7 @@ function parseLocale(str) {
     // handle the posix default locale
     if (str === "C") return "en-US";
 
+    // full locale
     if (str.length >= 10) {
         return [
             str.substring(0,2).toLowerCase(),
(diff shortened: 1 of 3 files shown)


js/lib/ilib.js
```diff
@@ -284,6 +284,43 @@ ilib.setLocale = function (spec) {
...
+    if (!str || str === "C") return str;
+
+    // take care of the libc style locale with a dot + script at the end
+    var dot = str.indexOf('.')
+    if (dot > -1) {
+        str = str.substring(0, dot);
+    }
+
+    // handle the posix default locale
+    if (str === "C") return "en-US";
```
gouniLee, line 300: It is already handled at line 291

js/lib/ilib.js
```diff
@@ -284,6 +284,43 @@ ilib.setLocale = function (spec) {
...
+        ].join("-");
+    }
+
+    if (str.length >= 5) {
+        return [
+            str.substring(0,2).toLowerCase(),
+            str.substring(3,5).toUpperCase()
+        ].join("-");
+    }
+
```
gouniLee, line 316: So we assume any of the platforms returns locale with language-script form.  never happen to get something like zh-Hans.  we get language-country or language-script-country form all the time. am I right?

ehoogerbeets: Addressed each comment. Please re-review!

gouniLee (approved):

PR #315, opened 2021-10-05
Support IoTjs

Seonghyup:
### Checklist

* [] At least one test case is included for this feature or bug fix.
* [] This is an API breaking change.
* [] Requires a major version change.

Support IoTjs which is something like lightweight js engine(or subset of nodejs).

### Links
https://iotjs.net/
https://jerryscript.net/

### DIFF
--- a/js/lib/ilib.js
+++ b/js/lib/ilib.js
@@ -137,8 +137,7 @@ ilib._getPlatform = function () {
                 return ilib._platform;
             }
         } catch (e) {}
-
-        if (typeof(global) !== 'undefined' && global.process && global.process.versions && global.process.versions.node && typeof(module) !== 'undefined') {
+	if (typeof(global) !== 'undefined' && global.process && ((global.process.versions && global.process.versions.node && typeof(module) !== 'undefined') || global.process.argv[0].includes("iotjs"))) {
             ilib._platform = "nodejs";
         } else if (typeof(Qt) !== 'undefined') {
             ilib._platform = "qt";


js/lib/ilib.js
```diff
@@ -137,8 +137,7 @@ ilib._getPlatform = function () {
                 return ilib._platform;
             }
         } catch (e) {}
-
-        if (typeof(global) !== 'undefined' && global.process && global.process.versions && global.process.versions.node && typeof(module) !== 'undefined') {
+	if (typeof(global) !== 'undefined' && global.process && ((global.process.versions && global.process.versions.node && typeof(module) !== 'undefined') || global.process.argv[0].includes("iotjs"))) {
```
ehoogerbeets, line 140: I think maybe using `typeof(process.iotjs) !== 'undefined'` might be a bit better because that way, it doesn't depend on the executable name including "iotjs" in it which could easily be renamed.

Also, the same change should be made to the https://github.com/ilib-js/ilib-env project.

I tried running the unit tests on iotjs and they failed because iotjs didn't support the "vm" module.

ehoogerbeets: Also, since nodejs and iotjs are not really the same platform, shouldn't we return a different platform string for it? That way, you can do different things based on the results. (Like maybe an IotJsLoader which is separate from the NodeLoader)

Seonghyup:
I will make another PR that check {{process.iotjs}} to support iotjs later.

PR #320, opened 2021-10-22
Add feature to set addtional path in Loader

gouniLee:
### Checklist

I've updated Loader to support multi-path. 
It is add new API add/remove additional path to Loader.
I've checked it works with sample node code. https://github.com/iLib-js/ilib-samples/pull/4
* In order to load data properly after setting paths, 
  * ilib data cached should be empty at this moment. if not, data is not loaded
  * set path order should be ordered first  (Even come first than  Resbundle's basePath option)

### Resolution

### Links


js/test/date/testMeridiems.js
```diff
@@ -2117,5 +2125,34 @@ module.exports.testmeridiems = {
...
+    testMeridiem_pa_IN_Custom: function(test) {
+        if (ilib._getPlatform() !== "nodejs" || !ilib._dyndata || !ilib._dyncode) {
+            test.done();
+            return;
+        }
+        var multiPath = path.relative(process.cwd(), path.resolve(__dirname, "./custom"));
+        var ilibLoader = ilib.getLoader();
+        
+        ilibLoader.addPath(multiPath);
+
```
ehoogerbeets, line 2147: Need a test for multiple multiPaths

js/lib/Loader.js
```diff
@@ -88,8 +88,16 @@ Loader.prototype._loadFileAlongIncludePath = function(includePath, pathname) {
 
 Loader.prototype.loadFiles = function(paths, sync, params, callback, root) {
     root = root || (params && params.base);
-    var includePath = root ? [root].concat(this.includePath) : this.includePath;
+    var includePath = this.includePath;
 
+    if (root) {
+        if (ilib._load.multiPaths){
+            includePath.splice(1, 0, root);
+        } else  {
```
ehoogerbeets, line 96: Before this change, we have:

1. root
2. includePaths (standard paths from the loader)

After this change, we have:

1. includePath[0] (the first one in the list, which is presumably the first multiPath)
2. root
3. includePath[1-X] (the second and subsequent multiPaths up to X)
4. includePath[X+1 - N] (the standard includePaths)

I think this is not what we want.

How about this:

this.includePath -> always the standard list that the loader defines
root -> always the root
this.multiPath -> any paths added/removed by the client using addPath() and removePath(). this.multiPath is initialized to the empty array [] in the constructor;

Then, the loadFiles should set:

```
var includePath = [].concat(this.multiPath); // makes a copy
if (root) includePath.push(root);
includePath.concat(this.includePath);

```

That way, we end up with:

1. All the multiPaths first
2. root
3. standard include paths from the loader

What do you think?

gouniLee: Codes would be more organized and neatly. I'll update it following your suggestion.

js/test/date/testMeridiems.js
gouniLee: Yeah, Actually I assume the added path would be one. I'll add more test cases while updating codes

js/lib/Loader.js
```diff
@@ -231,19 +229,23 @@ Loader.prototype.listAvailableFiles = function(sync, cb) {
...
 
-Loader.prototype.addPath = function (path) {
-    if (!path) return;
-    this.includePath.unshift(path);
-    this.multiPaths = true;
+Loader.prototype.addPath = function (paths) {
+    if (!paths) return;
+    paths = ilib.isArray(paths) ? paths : [paths];
+    this.addPaths = paths;
+    this.isMultiPaths = true;
```
ehoogerbeets, line 236: What if the client calls this:

```
loader.addPath("a/b");
loader.addPath("c/d/e/f");
```
Then, this would fail.

This method should be:

```
In the constructor:
this.addPaths = [];

In this method:
if (!paths) return;
var newpaths = ilib.isArray(paths) ? paths : [paths];
this.addPaths = this.addPaths.concat(newpaths);
```

The `this.isMultiPaths` does not seem to be used any more

gouniLee: I re-updated it. and in terms of  `this.isMultiPaths` is used here. https://github.com/iLib-js/iLib/blob/updateLoader/js/lib/Utils.js#L475

gouniLee:
@ehoogerbeets Is there anything else that I need to update more?

ehoogerbeets (approved):

gouniLee:
I've found additional issue case
if resources2/ko/strings.json 
{"hello": "안녕2"}

resources/ko/strings.json 
{"hello": "안녕",
 "thanks: "고마워"
}

In this case,  translated strings "thanks`"  returns source string "thanks"  in ko-KR locale.
Because it doesn't load resources/ko/strings.json file path because of line https://github.com/iLib-js/iLib/blob/development/js/lib/Loader.js#L74
so I update it more. Please re-review this change.

gouniLee:
@ehoogerbeets I updated codes more and added test cases. Please re-review them. Thank you.

gouniLee:
@ehoogerbeets  would it be ok to merge? and I need the newly released version which includes these changes  for the webOS submission

PR #324, opened 2021-12-09
Updated to route the code flow via NodeLoader when a platform is IoTjs

gouniLee:
### Checklist

It's related to https://github.com/iLib-js/iLib/pull/315 PR.
Updated code to route code flow via NodeLoader when a  platform is IoTjs. which is already applied to webos tv version ilib sources. 
Currently, iLib doesn't work on opensource version IoTjs  platform. but webOS TV is running on IoTjs  which are included many additional internal patches version.
Supporting IotNodeader might take loonng time. so Updated this code for the quick  fixes.

### Resolution

### Links
https://github.com/jerryscript-project/iotjs

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -13,4 +13,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-version=14.11.1
+version=14.11.2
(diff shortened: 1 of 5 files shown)


ehoogerbeets:
Unit test?

ehoogerbeets (approved):

PR #332, opened 2022-02-10
Update genclockprefs.js by reference different cldr and update clock.jf files

gouniLee:
### Checklist

Update genclockprefs.js by reference different cldr and update clock.jf files
Previously, clock.jf files are generated by reference `cldr-datas-full/main/[locale]/ca-gregorian.json`
but now it gets data from `cldr-core/supplemental/timeDate.json` which could give more correct data.

### Resolution

### Links


package.json
```diff
@@ -99,5 +99,8 @@
     },
     "publishConfig": {
         "tag": "experimental"
+    },
+    "dependencies": {
+        "ilib-resbundler": "^1.0.1"
```
ehoogerbeets, line 104: ilib should not have any direct dependencies

tools/cldr/genclockprefs.js
```diff
@@ -38,189 +37,62 @@ function usage() {
...
-        } else {
-            console.log("Skipping empty " + pathname);
-        }
-    } else {
-        console.log("Skipping existing " + pathname);
-    }
-}
-
-function getClockPrefs(locale) {
+for(var region in timeData){
```
ehoogerbeets, line 70: This is a locale, not a region. It took me a little time to realize that. Maybe rename the variable to "locale" ? Also maybe use `new Locale(locale)` to parse it instead of substrings?

js/test/date/testclock.js
```diff
@@ -1221,7 +1221,7 @@ module.exports.testclock = {
         var info = new LocaleInfo("en-CN");
         test.ok(info !== null);
 
-        test.equal(info.getClock(), 12);
+        test.equal(info.getClock(), 24);
```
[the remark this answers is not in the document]
ehoogerbeets: I vaguely remember the LG China office asked us to change it to 12 hour because of this...

ehoogerbeets (commented):
Please see the comments above

js/test/date/testclock.js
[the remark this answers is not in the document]
gouniLee: If cldr says, the default clock ks 24, shouldn't we follow it?It'w weird   how the current converting tool generate zh/clock.jf file as havine 12 so far. because cldr says 24.
If LG want to have different value than cldr, we can handle in custom locale data layer.

package.json
gouniLee: yes, it's shouldn't be there. I guess it's accidentally added. I updated it.

tools/cldr/genclockprefs.js
gouniLee: oh, I will rename the value as locale.  Most the data  in timeData.json is represent region so I thought not using `new Locale` would become code simpler  and doesn't need to have ilib dependency. If you think it's better to use new Locale for parsing locale, I'll update it too.

ehoogerbeets: There is common.Locale imported at the top of the file, so it doesn't have to rely on ilib itself

gouniLee: you are right. I updated the locale parsing part.

tools/cldr/genclockprefs.js
```diff
@@ -67,22 +67,25 @@ clockprefs = {
...
+    var lo = new Locale(locale);
+    var loStr=lo.getSpec();
+
+    if (loStr === "001") continue;
+    var clock = timeData[locale]["_preferred"];
     var outPath = "", fullPath = "";
 
-    if (Locale.isRegionCode(region)){
-        outPath = path.join("und", region);
+    if (Locale.isRegionCode(loStr)){
```
ehoogerbeets, line 82: `if (!lo.getLanguage()) {`

tools/cldr/genclockprefs.js
```diff
@@ -67,22 +67,25 @@ clockprefs = {
...
+        outPath = path.join("und", locale);
     } else {
-        if (region.indexOf("001") > -1){
-            outPath = region.slice(0, region.indexOf("001")-1);
+        if (loStr.indexOf("001") > -1){
```
ehoogerbeets, line 85: `if (lo.getRegion() === "001") {`

ehoogerbeets (approved):

PR #349, opened 2022-03-22
Update list.json  for having sorted data and related modification

gouniLee:
### Checklist

* Update `list.json` to have sorted data.
* Fixed some tool code to work properly.
* Fix tc failures which were wrong  applied cldr 40 changes.

### Resolution

### Links


js/test/strings-ext/testlistfmt.js
```diff
@@ -2825,7 +2825,7 @@ module.exports.testlistfmt = {
 
         test.expect(2);
         test.ok(fmt !== null);
-        test.equal(fmt.format(["một", "hai"]), "một và hai");
+        test.equal(fmt.format(["một", "hai"]), "một, hai");
```
ehoogerbeets, line 2828: Why did this change? The vi/list.json is re-arranged, but has the exact same content as before...

tools/cldr/genlist.js
```diff
@@ -132,20 +132,20 @@ var hardCodeData = {
     "ig":{
         "or": {
             "short": {
-                "2": "{0} mba {1}",
-                "end": "{0}, mba {1}"
+                "2": "{0} ma ọ bụ {1}",
```
ehoogerbeets, line 135: How did this change, but there are no unit test updates?

js/test/strings-ext/testlistfmt.js
gouniLee: It's weird to me too. I just re-ran genlist.js and get the output. and newly generated output is correct with comparing cldr 40.I assume it might be missed when doing cldr 40 update work.

gouniLee: I ran the `genlsit.js` tool code with not supporting sorting feature. that is the same result that I currently get. 
It's obvious not correctly applied cldr40 changes before.

here's cldr40 result in vi/sysres.json 
```
"unit": {
        "short": {
            "2": "{0} {1}",
            "start": "{0} {1}",
            "middle": "{0} {1}",
            "end": "{0} {1}"
        }
    }
```
current data:  vi/list.json
```
"unit": {
        "short": {
            "2": "{0} {1}",
            "start": "{0} {1}",
            "middle": "{0} {1}",
            "end": "{0} {1}"
        },
        "medium": {
            "2": "{0} và {1}",
            "end": "{0} và {1}"
        },
        "long": {
            "2": "{0} và {1}",
            "end": "{0} và {1}"
        },
        "full": {
            "2": "{0} và {1}",
            "end": "{0} và {1}"
        }
    }
```
`

ehoogerbeets (approved):

PR #353, opened 2022-04-06
Update ReleaseNotes for the previous PR

gouniLee:
### Checklist

Update ReleaseNotes and version number for the next release

### Resolution

### Links

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -13,4 +13,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-version=14.15.0
+version=14.14.1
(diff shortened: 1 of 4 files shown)


build.properties
```diff
@@ -13,4 +13,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-version=14.14.1
+version=14.15.0
```
ehoogerbeets, line 16: The minor version should change when a new feature is added. In this case, we are just rearranging some json files and fixing a few CLDR generation problems along the way, so I think that is a point release only. Please set this back to 14.14.1

ehoogerbeets (approved):

PR #354, opened 2022-04-08
CLDR 41 Update - date formats

ehoogerbeets:
### Checklist

### Resolution

### Links


ehoogerbeets:
@gouniLee This one should go in first. After it is merged, we can rebase the other ones to get a much smaller set of files and make it much easier to code review.

tools/cldr/gencurrencies.js
```diff
@@ -111,7 +111,7 @@ if (process.argv.length < 2) {
 }
 
 if (process.argv.length > 2) {
-    ilibDir = process.argv[2];
```
gouniLee, line 114: This change might give confusion. 
Is for referencing ilib location while generating data. It's not for output location.

When running gencurencies.js, the ilib locale data path should be put as a parameter mandatory. because if the cldr doesn't have proper data, we've decided to maintain current ilib data.
Usage: gencurrency [iLib locale Dir] [toDir]

gouniLee (commented):
doesn't need to add additional test cases regarding locale data change?

ehoogerbeets:
> doesn't need to add additional test cases regarding locale data change?

CLDR 41 did not change much in the existing locales. Mostly, it added a few new locales that we don't test.

tools/cldr/gencurrencies.js
ehoogerbeets: put it back the way it was, but added a declaration for the variable that node complained about

gouniLee (approved):

PR #359, opened 2022-05-04
Added option to choose whether to use Intl.DateTimeFormat

gouniLee:
### Checklist

Added option to choose whether to use Intl Obj. If the user could set `useIntl:true` when creating DateFmt object. then options are available for Intl Object, it uses it for formatting. 
If iLib use IntlObject, it could result in better speed, but iLib does not handle the output at all. the result may differ depending on the platform and version 

### Resolution

### Links

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -8,6 +8,7 @@ Published as version 14.15.0
 New Features:
 * Update to CLDR v41 data
 * Update to the latest ISO-14924 data (writing script information)
+* Added `useIntl` option in DateFmt to choose whether to use Intl.DateTimeFormat
 
 Bug Fixes:
 * Fixed a bug where the DateFmt.formatRelative() does not represent correct result in certain case.
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -9,6 +9,7 @@ New Features:
 * Update to CLDR v41 data
 * Update to the latest ISO-14924 data (writing script information)
 * Added `useIntl` option in DateFmt to choose whether to use Intl.DateTimeFormat
+    * When it is is true, and paramaters are convertable for using Intl Object case, you can get relatively fast result, but different result may get depending on the platform nd version.
 
 Bug Fixes:
 * Fixed a bug where the DateFmt.formatRelative() does not represent correct result in certain case.
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -9,7 +9,10 @@ New Features:
 * Update to CLDR v41 data
 * Update to the latest ISO-14924 data (writing script information)
 * Added `useIntl` option in DateFmt to choose whether to use Intl.DateTimeFormat
-    * When it is is true, and paramaters are convertable for using Intl Object case, you can get relatively fast result, but different result may get depending on the platform nd version.
+    * When it is set to true, the Intl object is available, it supports the requested locale, and the parameters can be converted to equivalent parameters for the Intl.DateTimeFormat object, then it will format the date relatively quickly using Intl.
+    * When they cannot be converted, the Intl object is not available, or the Intl object does not support the requested locale, it will perform the relatively slow formatting using regular ilib code written in Javascript.
+    *  The code will often return different results depending on the platform and version of the Javascript engine and which version of CLDR it supports. If you need consistency across versions and platforms, do not use the useIntl flag. Just stick with the regular ilib formatting code.
+
 
 Bug Fixes:
 * Fixed a bug where the DateFmt.formatRelative() does not represent correct result in certain case.
--- a/js/lib/DateFactory.js
+++ b/js/lib/DateFactory.js
@@ -259,4 +259,19 @@ DateFactory._dateToIlib = function(inDate, timezone, locale) {
     });
 };
 
+DateFactory._ilibToDate = function(ilibDate, timezone, locale) {
+    if (typeof(ilibDate) === 'undefined' || ilibDate === null) {
+        return ilibDate;
+    }
+    var year = ilibDate.year;
+    var month = ilibDate.month-1;
+    var day = ilibDate.day;
+    var hours = ilibDate.hour;
+    var minutes = ilibDate.minute;
+    var seconds = ilibDate.second;
+    var milliseconds = ilibDate.second;
+
+    return new Date(year, month, day, hours, minutes, seconds, milliseconds);
+};
+
 module.exports = DateFactory;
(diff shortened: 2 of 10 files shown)


js/lib/DateFactory.js
```diff
@@ -259,4 +259,19 @@ DateFactory._dateToIlib = function(inDate, timezone, locale) {
     });
 };
 
+DateFactory._ilibToDate = function(ilibDate, timezone, locale) {
+    if (typeof(ilibDate) === 'undefined' || ilibDate === null) {
```
ehoogerbeets, line 263: ilibDate could be:

- a number (unix time)
- a string (ISO 8601 date format)
- an intrinsic Javascript Date object
- null, undefined, boolean, Symbol, array, or object other than an IDate
 
So this should catch all possibilities:

```
if (typeof(ilibDate) !== 'object' || !(ilibDate instanceof IDate)) {
    return ilibDate;
}
```

js/lib/DateFactory.js
```diff
@@ -259,4 +259,19 @@ DateFactory._dateToIlib = function(inDate, timezone, locale) {
...
+        return ilibDate;
+    }
+    var year = ilibDate.year;
+    var month = ilibDate.month-1;
+    var day = ilibDate.day;
+    var hours = ilibDate.hour;
```
ehoogerbeets, line 269: `new Date()` uses the local time zone offset only. So if the ilibDate does not use the "local" time zone and it does not use the current time zone, then the above doesn't work. We can always get the right time by doing this:

```
return new Date(ilibDate.getTimeExtended());
```

`ilibDate.getTimeExtended()` returns the UTC-based unix time, and `new Date` accepts a single parameter like that. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date

js/lib/DateFmt.js
```diff
@@ -437,60 +451,75 @@ var DateFmt = function(options) {
...
             this.calName = this.calName || this.locinfo.getCalendar() || "gregorian";
-            if (ilib.isDynCode()) {
-                // If we are running in the dynamic code loading assembly of ilib, the following
-                // will attempt to dynamically load the calendar date class for this calendar. If
-                // it doesn't work, this just goes on and it will use Gregorian instead.
-                DateFactory._init(this.calName);
+            
+            if(sync && typeof(Intl) !== 'undefined' && Intl.DateTimeFormat.supportedLocalesOf(this.locale.getSpec()).length > 0){
+                if (this.locinfo.getDigitsStyle() == "western" && (!options.template) && this.calName == "gregorian")    {
+                    if(this.type == "date" && ((this.dateComponents == "dmy" && this.length != "f") ||
```
ehoogerbeets, line 457: All three lines can be one long "if" condition.

Also, we don't need to check for sync. We can create the Intl date formatter synchronously and then call the callback as if it were asynchronous.

`this.type == "date"` -> we can also format times with the Intl formatter, can't we?

gouniLee: Yes, it is a middle of progress that haven't written completely. I just made a draft PR for the preparing discussion our meeting this week (Unfortunately we missed it). I'm not sure if you had chance to read mail that I send to you last Tuesday, I wrote some issues regarding the implementation these way.  Let's discuss these in next meeting and then will continue to work on.

js/test/date/testdatefmtasync.js
```diff
@@ -304,5 +304,166 @@ module.exports.testdatefmtasync = {
...
+                    locale: "en-US",
+                    useIntl: true,
+                    sync: false,
+                    onLoad: function(fmt){
+                        if(ilib._getPlatform() === "nodejs"){
+                            var version = process.versions["node"];
+                            var majorVersion = version.split(".")[0];
+                            //console.log("version: " + version);
+                            if(majorVersion == "8" || majorVersion == "10"){
+                                test.equal(fmt.format(DateFactory._ilibToDate(date)), "9/29/2022");
```
ehoogerbeets, line 426: `format()` already calls `DateFactory._ilibToDate()` doesn't it?

ehoogerbeets: Okay just checked. It does. You should be able to do `fmt.format(date)`

js/test/date/testdatefmt.js
```diff
@@ -3981,5 +3981,12 @@ module.exports.testdatefmt = {
 
         test.equal(fmt.getDateComponentOrder(), "mdy");
         test.done();
+    },
+    testDateFmtisIntlDateTimeAvaiable: function(test) {
+        test.expect(1);
+
+        var result = DateFmt.isIntlDateTimeAvailable();
+        test.equal(result, false);
```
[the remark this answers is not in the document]
gouniLee: This case is for testing there's no parameter. I intended to return `false` if there's no parameter.

js/test/date/testdatefmtasync.js
gouniLee: In this case, it returns Intl's object. so this format method is for Intl's one. not iLib one.  It should take js date, instead iLib date style.

ehoogerbeets: Oh wait, that's a problem. The "useIntl" parameter is a request to use Intl if it is available. If it is not available, or if the parameters cannot be mapped to Intl, or if the locale is not support, then the format object returned is the ilib object. If Intl is available, then it is the Intl object. The caller won't know until they check the object. Also, it is weird for a constructor to return an object of a different type than itself.

It would be better to change the code so that instead of returning the Intl formatter, you always return the ilib formatter:

```javascript
constructor:
if (useIntl) { this.intlFormatter = Intl.dateTimeFormat(); return; }

format:
if (useIntl) { convert date to JS date; return this.intlFormatter(JS Date); }
```

gouniLee: I updated it. Please re-check it.

docs/ReleaseNotes.md
```diff
@@ -8,6 +8,8 @@ Published as version 14.15.0
 New Features:
 * Update to CLDR v41 data
 * Update to the latest ISO-14924 data (writing script information)
+* Added `useIntl` option in DateFmt to choose whether to use Intl.DateTimeFormat
+    * When it is is true, and paramaters are convertable for using Intl Object case, you can get relatively fast result, but different result may get depending on the platform nd version.
```
ehoogerbeets, line 12: Should be:

- When it is set to true, the Intl object is available, it supports the requested locale, and the parameters can be converted to equivalent parameters for the Intl.DateTimeFormat object, then it will format the date relatively quickly using Intl.
- When they cannot be converted, the Intl object is not available, or the Intl object does not support the requested locale, it will perform the relatively slow formatting using regular ilib code written in Javascript.
- The code will often return different results depending on the platform and version of the Javascript engine and which version of CLDR it supports. If you need consistency across versions and platforms, do not use the useIntl flag. Just stick with the regular ilib formatting code.

ehoogerbeets (approved):

PR #366, opened 2022-06-28
Add hardCoded locales for localematch

gouniLee:
### Checklist

Add hardCoded locales for localematch.json
Other changes are due to re-running genlikelyloc.js with cldr 41.

### Resolution

### Links


gouniLee:
> This same change needs to go into ilib-localematch as well.

I checked the localematch.json in ilib-localematch. there are many differences already. It looks like it hasn't applied the latest changes.

PR #367, opened 2022-06-28
Adopt Intl.Locale in LocaleInfo

gouniLee:
### Checklist

The idea is to use Intl.Locale if it's available in the constructor. then In the method, make to load locale data if it's needed.
**The synchronous call is only available.** 

Currently, ` getScript()`, `getFirstDayOfWeek()`, `getWeekEndStart()` , `getWeekendEnd()`, `getClock()` are directly use Intl.Locale Information. there's no conflict between iLib and Intl.Locale data if the input locale is valid.
currently, exceptional locales are for testing locales for test cases and when custom overriding data is exist.

Checked various node versions (Intl.Locale is valid over `node v12.0`)
Checked `chrome(v103.0.5060,53)` and `firefox(v101.0.1)` browsers.

### Resolution

### Links
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale


gouniLee:
discussed this approach, and decided it's not an appropriate idea. so closing this PR.

PR #375, opened 2022-07-26
CLDR 41 Update - list

gouniLee:
### Checklist

* Fix `genlist.js` tool code 
  *  Remove promoteFormats() which in inappropriate cause it override fallback locale data (i.e. en-CA data overrides en/ data due to that line )
*  Re-generate `list.json` files and fixed related test cases which were wrong.

### Resolution

### Links


tools/cldr/genlist.js
```diff
@@ -321,14 +321,6 @@ console.log("\n\nMerging formats forward ...");
 
 aux.mergeFormats(localePatterns, localePatterns, []);
 
-console.log("\n\nPromoting sublocales ...");
```
ehoogerbeets, line 324: Why is this removed?

gouniLee: I realized the data have generated wrongly due to this line. 
It overrides fallback locale data. for example, en-CA data overrides en data which should maintain it.

ehoogerbeets (approved):

PR #379, opened 2022-08-05
 Fixed to generate `plurals.json` files when the rule only has `other`

gouniLee:
### Checklist

* Fixed to generate `plurals.json` files even when the rule only has `other`
* Fixed some wrong locale match with test function name (It seems to happened when multiple branches were merged at the same time when I was working to add new locales)

Issue case example)
Korean has no plural rules (only `other` exists.) so, genplural tool code doesn't generate plurals.json file for korean. If no plurals.json file to load iLIb applies the default rule which is the same is en-US that has (one-other). so In `one` case for koerean , iLib returns `one` category but it should `other`.

### Resolution

### Links


docs/ReleaseNotes.md
```diff
@@ -1,6 +1,14 @@
 Release Notes for Version 14
 ============================
 
+Build 023
+-------
+Published as version 14.16.0
```
ehoogerbeets, line 6: For semantic versioning, a bug fix should only affect the 3rd level of version. So, this should be v14.15.1.

But, having said that, we haven't shipped v14.15.0 yet because of the unit test failure, so you can probably add this bug fix to that release instead.

ehoogerbeets (approved):

PR #383, opened 2022-08-09
Updated to load plurals.json in ResBundle Constructor

gouniLee:
### Checklist

Updated to load `plurals.json` in ResBundle Constructor, then you could call call synchronously all the time because we can be sure if is already loaded.

### Resolution

### Links


docs/ReleaseNotes.md
```diff
@@ -1,6 +1,14 @@
 Release Notes for Version 14
 ============================
 
+Build 024
+-------
```
ehoogerbeets, line 5: We haven't shipped 14.15.0 yet, so this should be part of that release

ehoogerbeets (approved):

PR #384, opened 2022-08-10
Remove json files that have no contents

gouniLee:
### Checklist

Remove `dateformats.json` and `sysres.json` files that have no contents. 
I assume it's a legacy. these files are not generated when running cldr tool code (`gendatefmt2.js`)

### Resolution

### Links


ehoogerbeets (approved):
We should also modify the gendatefmts2.js tool so that it never generates empty files

gouniLee:
> We should also modify the gendatefmts2.js tool so that it never generates empty files

As I wrote in the description, When I'm checking, It doesn't generate these files by running `gendatefmt2.js` 

(fyi.  when I'm running gendatefmt2.js tool code with cldr41. I found there are some differences between my output and current data) It should be no diff because you already did. It's weird. need to figure out the reason..)

PR #388, opened 2022-08-19
Fixed to handle a path correctly in ilib-web.js

gouniLee:
### Checklist

Fixed to handle both absolute and relative paths properly in `ilib-web.js`
(`ilib-web.js` file is still used in webOS Signage.)

### Resolution

### Links

### DIFF
--- a/js/lib/ilib-web.js
+++ b/js/lib/ilib-web.js
@@ -59,9 +59,9 @@ var requireClass = function() {
     for (var i = 0; i < scripts.length; i++) {
         var source = scripts[i].src;
         if (source && (pos = source.search(/ilib-web\.js$/)) !== -1) {
-            var colon = source.indexOf('://');
-            this.protocol = source.substring(0,colon+3);
-            this.root = source.substring(colon+3, pos-1);
+            var url = new URL(source);
+            this.protocol = url.protocol + "//";
+            this.root = url.pathname;
             break;
         }
     }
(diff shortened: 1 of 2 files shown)


js/lib/ilib-web.js
```diff
@@ -59,9 +59,9 @@ var requireClass = function() {
     for (var i = 0; i < scripts.length; i++) {
         var source = scripts[i].src;
         if (source && (pos = source.search(/ilib-web\.js$/)) !== -1) {
-            var colon = source.indexOf('://');
-            this.protocol = source.substring(0,colon+3);
-            this.root = source.substring(colon+3, pos-1);
+            var url = new URL(source);
```
ehoogerbeets, line 62: URL doesn't work on IE: https://developer.mozilla.org/en-US/docs/Web/API/URL

js/lib/ilib-web.js
```diff
@@ -58,11 +58,23 @@ var requireClass = function() {
 
     for (var i = 0; i < scripts.length; i++) {
         var source = scripts[i].src;
-        if (source && (pos = source.search(/ilib-web\.js$/)) !== -1) {
-            var url = new URL(source);
-            this.protocol = url.protocol + "//";
-            this.root = url.pathname;
+
+        if (navigator && navigator.userAgent && (navigator.userAgent.indexOf(" .NET") > -1) ){
+            // IE brower
```
ehoogerbeets, line 63: This should use ilib._getBrowser() instead of figuring it out on your own.

gouniLee: I guess this part works before requiring ilib codes.

ehoogerbeets: Oh I see... so that means ilib._getBrowser() is not available yet. Okay

ehoogerbeets (approved):

PR #399, opened 2022-10-30
Updated to UCD v15.0.0

ehoogerbeets:
### Checklist

### Resolution

### Links


build.xml
```diff
@@ -115,6 +115,10 @@ limitations under the License.
         <mkdir dir="${build.dist}" description="Dir to contain all the distribution outputs"/>
     </target>
 
+    <target name="test.dynamic.uncompiled">
```
gouniLee, line 118: Is there any special reason to change to `test.dynamic.uncompiled` instead of `ant test`?

ehoogerbeets: I was using the tool `npm-check-updates` to update/upgrade packages. This tool runs the "test" script in the package.json, then attempts to upgrade everything, and then runs the "test" script again to see that nothing is broken. If it is broken, then it backs off various upgraded packages one at a time until it does work again. This means it might run the "test" script many times. Currently, `ant test` takes 15 minutes, whereas `test.dynamic.compiled` takes only 3 or so on my machine. The target `test.dynamic.compiled` is good enough for npm-check-updates to check the package updates and takes way less time.

gouniLee (approved):

PR #419, opened 2023-02-14
CLDR 42 Update - langname

gouniLee:
### Checklist

* Updated `langname.jf` to cldr42 by running `genlangreg.js`

### Resolution

### Links

### DIFF
--- a/js/data/locale/root/langname.jf
+++ b/js/data/locale/root/langname.jf
@@ -1,4 +1,0 @@
-{
-    "generated": true,
-    "language.name": "Root"
-}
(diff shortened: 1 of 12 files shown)


js/data/locale/root/langname.jf
```diff
@@ -1,4 +0,0 @@
-{
-    "generated": true,
-    "language.name": "Root"
-}
```
ehoogerbeets, line 4: I don't think this one should be removed. If the locale has no language tag or the language tag does not have a langname.jf, then this one takes over....

gouniLee: then, it should be located under locale/ directory, not locale/root/ directory. no?

ehoogerbeets: Ah yes, I missed that part.

ehoogerbeets (approved):

PR #430, opened 2023-07-04
Add more pseudomap.json data files

gouniLee:
### Checklist

* Added additional `pseudomap.json` data file.  it's related https://github.com/iLib-js/loctool/pull/225
It return pseudo string when the locale is set as below.
`am-XX,as-XX,bn-XX,hi-XX,mr-XX,gu-XX,pa-XX,kn-XX,ml-XX,or-XX,ta-XX,te-XX,km-XX,si-XX,ka-XX`
*  Added tall properly in some pseudo map data which is the tallest glyphs in the script  (did search as much we could, but may not accurate)  it is going to be included first in a pseudo string to see if it doesn't break the UI.

### Resolution

### Links


docs/ReleaseNotes.md
```diff
@@ -7,6 +7,9 @@ Published as version 14.18.0
 
 New Features:
 * Updated pseudo strings to be encapsulated with square brackets so that you can tell when the resource string starts and ends.
+* Added more pseudomap data to test by adding pseudoLocale.
+    * Added tall properly in some `pseudomap.json` files which is the tallest glyphs in the script
+      and it is going to be included first in a pseudo string.
```
ehoogerbeets, line 12: * Added more pseudomap data for use in testing different fonts using pseudo locales.
    * Added a "tall" property in some `pseudomap.json` files which gives the tallest glyph in the script.
      This tallest glyph is included at the beginning of a pseudo localized string in order to test
      if there is enough room for the height of text in each UI element.

ehoogerbeets (approved):

docs/ReleaseNotes.md
gouniLee: Thank you! :)

PR #441, opened 2023-10-11
CLDR 42 Re-Update - Country with data sorting

gouniLee:
### Checklist

* [ ] At least one test case is included for this feature or bug fix.
* [ ] `ReleaseNotes` has added or is not needed.
* [ ] This PR has passed all of the test cases on `Nodejs` and `Chrome Browser`.
* [ ] This PR has passed all of the test cases on `QT/QML`.
* [ ] This is an API breaking change.
* [ ] Requires a major version change.

### Issue Resolved / Feature Added
[//]: # (Describe the issue resolved or feature added by this pull request)

### Resolution
I found the `regionnames.json` file were not update when I was working cldr v42 update. and the tool code (`gencountrynames.js`) haven't been sorting data by using 'json-stable-stringify`. so I updated tool code and update output agagin base on v42. 
then Update cldr v43 work will do again based on this change.

### Links
[//]: # (Related issues, references)

PR #447, opened 2023-10-30
Update CLDR 43.1 - Country

gouniLee:
### Checklist

Update ctrynames.js, ctryreverse.json and regionnames.json to cldr v43.1
https://github.com/iLib-js/iLib/pull/441 PR has to be merged first then, we can check the changes only made in cldr v 43.1
major changes: 
 - new country data is added : CQ
 - TR contryname has changed : Türkiye

### Resolution

### Links


js/data/locale/ar/BH/ctrynames.json
```diff
@@ -292,5 +292,6 @@
     "هولندا الكاريبية": "BQ",
     "هونغ كونغ الصينية (منطقة إدارية خاصة)": "HK",
     "وسط آسيا": "143",
-    "وسط أفريقيا": "017"
+    "وسط أفريقيا": "017",
+    "Sark": "CQ"
```
ehoogerbeets, line 296: This is not right... Sark has the ISO 3166 code of GG, not CQ. CQ is the Northern Marianas Islands.

Also, the name does not seem to be translated to any of the languages in any of these files.

gouniLee: But, this data is newly added from cldr v43. from cldr-localenames-full package.
i.e) https://github.com/unicode-org/cldr-json/blame/main/cldr-json/cldr-localenames-full/main/en-AE/territories.json
Do you think it better filter the data out in the tool code?

gouniLee: and the reason the data has not translated case, it's because of the tool code. the tool code does something.
it looks like country code information `CQ` is only added `en-Region` locale and `hi-Latn` only.

gouniLee: Regarding the code `CQ`
https://www.iso.org/obp/ui/#iso:code:3166:CQ
> "Previously Sark was represented by Guernsey's country code (GG). Sark also fought for 20 years to get the .cq country code top-level domain."
from Politics https://en.wikipedia.org/wiki/Sark

the chromium has been updated too. https://chromium-review.googlesource.com/c/chromium/src/+/4783452

and It looks like MP is Northern Marianas Island 's ISO code.  (https://en.wikipedia.org/wiki/Northern_Mariana_Islands)

ehoogerbeets (approved):

PR #456, opened 2024-01-29
CLDR 44.1 Update - LocaleMatch, langname

gouniLee:
### Checklist

* Update localematch.json and langname.jf files  to cldr v44.1 
    * No changes to regionname.jf files
* Add some test cases
* Remove invalid test cases that are no longer valid

### Resolution

### Links


js/data/locale/localematch.json
```diff
@@ -1,35 +1,7 @@
 {
     "likelyLocales": {
-        "002": "en-Latn-NG",
-        "003": "en-Latn-US",
-        "005": "pt-Latn-BR",
-        "009": "en-Latn-AU",
```
ehoogerbeets, line 6: What happened to these numbered region? They should still be there.

gouniLee: There are the chages from v43.1. Many of und-002, und-053 locales are gone in v44.1 (cldr-core/supplemental/likelySubtags.json)
Should we keep these ? We can add them in the tool code.

gouniLee: I added these deleted locales to the tool code and re-generated the localematch.json file.

ehoogerbeets (approved):

PR #463, opened 2024-02-01
Update to IANA timezone 2024a data

ehoogerbeets:
### Checklist

### Resolution

### Links


js/test/calendar/testpersiandateastro.js
```diff
@@ -620,7 +620,7 @@ module.exports.testpersiandateastro = {
 
         test.ok(pd !== null);
 
-        test.equal(pd.getJulianDay(), 2458201.1965277777);
+        test.equal(pd.getJulianDay(), 2458201.238194444);
```
gouniLee, line 623: Did you calculate by yourself? How did this number come out?

ehoogerbeets: Iran ended DST this year, and their DST offset was 1 hour. One hour in Julian days is 0.041666666666666664 (this is 1/24 of a day). So 2458201.1965277777 + 0.041666666666666664 = 2458201.238194444  (to 16 decimal places -- close enough!).

ehoogerbeets: The reason it is plus one hour is that the test calculates the julian day from the year, month, day, etc. The date is the first month, third day which was in daylight time before (+05). Now that same date is in standard time (+06 as it is all year long), so 1 hour later than before.

gouniLee: I see. Thanks for the explanation! By the way, Iran DST Info changed in 2022b. I think it should have been changed earlier. isn't it? https://mm.icann.org/pipermail/tz-announce/2022-August/000071.html

gouniLee (approved):

js/test/calendar/testpersiandateastro.js
ehoogerbeets: Actually, I had it wrong. It was Kazakhstan that changed from +05 to +06. Iran changed from +3:30/+4:30 to just +3:30 only. I'm not sure what happened between 2022b and now, but the IANA data says that Iran stopped DST just now, not in 2022. We can only go with what they say because we don't know any better than them.

PR #483, opened 2025-03-18
Add ko-TW locale to localematch.json

gouniLee:
### Checklist

Add `ko-TW` locale to localematch.json that is currently returning ko-Kore-KR instead of ko-Kore-TW
note) The ko-TW locale has recently been added to webOS as a valid locale from a localization perspective.

### Resolution

### Links


ehoogerbeets (approved):
How about the same change in ilib-localematcher?

gouniLee:
> How about the same change in ilib-localematcher?

Yes. I created the PR for that. https://github.com/iLib-js/ilib-localematcher/pull/12
However, CircleCI throws an error during the installation of Firefox, which could not be reproduced locally.

PR #492, opened 2026-04-08
Fixes a bug in `LocaleMatcher` where locale parsing information is passed incorrectly

gouniLee:
### Checklist

Align the variable names passed by Localematcher with the order expected by the actual Locale object to reduce confusion.
This change is not a bug and does not affect existing behavior.
```
// Locale.js
var Locale = function(language, region, variant, script) {
...
```

### Resolution

### Links


js/lib/LocaleMatcher.js
```diff
@@ -144,22 +144,22 @@ LocaleMatcher.prototype = {
 
         if (typeof(this.info.likelyLocales[locale.getSpec()]) === 'undefined') {
             // try various partials before giving up
-            var partial = this.info.likelyLocales[new Locale(locale.language, undefined, locale.region).getSpec()];
+            var partial = this.info.likelyLocales[new Locale(locale.language, locale.region, undefined, undefined).getSpec()];
```
ehoogerbeets, line 147: Um, not sure how this could NOT be a bug! Can you add some unit tests to test for this?

gouniLee: Ah, you're right.. It is a bug. I confused it with the part that parses the first argument. I’ve added a test case. Please take a look. I will apply the same related code updates to ilib-localematcher as well.

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,16 @@
 Release Notes for Version 14
 ============================
 
+Build 034
+-------
+Published as version 14.22.0
```
ehoogerbeets, line 6: This is a bug fix, so it should be 14.21.4

js/test/root/testlocale.js
```diff
@@ -392,6 +392,43 @@ module.exports.testlocale = {
         test.ok(typeof(loc.getVariant()) === "undefined");
         test.done();
     },
+    testLocaleConstructorPartial2: function(test) {
+        test.expect(5);
```
ehoogerbeets, line 396: This tests the locale constructor itself, but the code that changed was the LocaleMatcher class. Please add LocaleMatcher tests where the whole locale is unknown, but various partial locales match.

gouniLee: I have added the test cases. actually, similar test cases already exist in LocaleMatcher.

docs/ReleaseNotes.md
gouniLee: That's correct. However, since I knew it wouldn't be released as 14.21.4, I just wrote it as 22.0. I updated it to 21.4. later  I'll fix it when there's a conflict in the future.

js/test/root/testlocalematch.js
```diff
@@ -2832,5 +2832,60 @@ module.exports.testlocalematch = {
         test.ok(typeof(locale) !== "undefined");
         test.equal(locale.getSpec(), "ko-CN");
         test.done();
+    },
+    testLocaleMatcherGetLikelyLocaleMinimalNonDefaultLocale_ko_CN: function(test) {
```
ehoogerbeets: wrong test name: 
```suggestion
    testLocaleMatcherGetLikelyLocaleMinimalNonDefaultLocale_ar_Arab_EG: function(test) {
```

ehoogerbeets (approved):

PR #493, opened 2026-04-09
CLDR 48.2 Update - Number, Currency and Delimeters

seonmiJin:
### Checklist

* Update `number.jf` to CLDR v48.2 by running `gennumfmt.js`
  * Although CLDR 48 defines **ku-Arab** as Kurdish (Kurmanji, Arabic script),
  the semantic meaning of ku-Arab-IQ in webOS may vary depending on its historical adoption,
  given that some earlier systems used **ku-Arab-IQ** as a proxy for **Central Kurdish (ckb)**.
* Update `currency.json` and `currency.js` by running `gencurrencies.js`
  * Update `gencurrencies.js` to sort ascending by _from date so the most recent currency is last.
* Update `delimeters.jf` by running `gendelimiters.js`
* Add and update related test cases

### Resolution

### Links


ehoogerbeets (approved):

js/data/locale/und/BG/currency.jf
```diff
@@ -1,3 +1,3 @@
 {
-    "currency": "BGN"
```
gouniLee, line 2: The NA/BT/NA/currency.jf files appear to be correct based on the previous data. Could you please check them again?

seonmiJin: A bug in the tool code was fixed by sorting by _from date in ascending order, and the data was regenerated. BT, NA, and ZW remain unchanged from the previous version. Note that BG (Bulgaria) now uses EUR as its default currency.

gouniLee (approved):

PR #496, opened 2026-04-21
CLDR 48.2 Update - Date, ClockPrefs

seonmiJin:
### Checklist

* Note: The reason this PR includes a large number of changed files is that during the CLDR v46 sync, values identical to those in *root* were not pruned and were instead included in each locale’s data.
* Update `dateformats.json` to CLDR v48.2 by running `gendatefms2.js` **except ku locales**
  * gendatefms2: Skip **cop** and **ku** locales
  * datefmts.js: Handle zha timezone-hour-am/pm order (ex. ml-IN)
* Update `clock.jf`  to CLDR v48.2 by running `genclockprefs.js`
* Add **en-JP** to locales.unittest
* Update affected unit tests

### Resolution

### Links


gouniLee (approved):

gouniLee:
@ehoogerbeets Could you please review this PR? In order to work on the Kurdish-related tasks we discussed in the last meeting in a separate PR, this PR needs to be merged first.

tools/cldr/datefmts.js
```diff
@@ -954,6 +958,22 @@ module.exports = {
                             i = i < begin ? end : i;
                             zTemplate = "{time}" + longtime.substring(i);
                             break;
+                        case 'zha':
+                            begin = scanForLastChars(shorttime, "m");
+                            end = scanForChars(shorttime, "a");
+                            i = end;
+                            while (shorttime.charAt(i) !== ' ' && shorttime.charAt(i) !== ' ' && i > begin) {
```
ehoogerbeets: This tests against the regular ascii space character twice. This was probably copied from line 944 or 955, which use one ascii space and the other is a narrow non-breaking space (U+202F). It is confusing because it is hard to see in the editor!

```suggestion
                            while (shorttime.charAt(i) !== ' ' && shorttime.charAt(i) !== '\u202F' && i > begin) {
```

If you take that suggestion, please update the other similar lines above as well so that it is less confusing for us all in the future.

tools/cldr/gendatefmts2.js
```diff
@@ -298,7 +298,7 @@ aux.mergeFormats(systemResources, systemResources, []);
 console.log("\nReading CLDR data ...");
 
 //these locales have the wrong data in CLDR and need to be skipped for now
-var skipList = ["ku"];
+var skipList = ["cop", "ku"];
```
ehoogerbeets, line 301: Why is "cop" added to the skip list?

js/test/durfmt/testdurfmtasync.js
```diff
@@ -149,7 +149,7 @@ module.exports.testdurfmtasync = {
                     second: 1,
                     millisecond: 1
                 });
-                test.equal(duration.toString(), '1 J, 1 M, 1 W, 1 T, 1 Std., 1 Min., 1 Sek., 1 ms');
+                test.equal(duration.toString(), '1 J, 1 M, 1 W, 1 T, 1h, 1 Min., 1 Sek., 1ms');
```
ehoogerbeets, line 152: Are we sure this is right? German for "hour" is "Stunde", so "h" doesn't seem right...

seonmiJin: In CLDR 48, the narrow unit data has been updated as follows.

cldr-units-full/main/de/units.json
```
// cldr47
"duration-hour": {
  "displayName": "Std.",
  "unitPattern-count-one": "{0} Std.",
  "unitPattern-count-other": "{0} Std.",
  "perUnitPattern": "{0}/h"
}
// cldr48
"duration-hour": {
  "displayName": "Std.",
  "unitPattern-count-one": "{0}h",
  "unitPattern-count-other": "{0}h",
  "perUnitPattern": "{0}/h"
}
```

tools/cldr/gendatefmts2.js
seonmiJin: The data for "cop" was manually authored and has not been generated automatically.
To reflect this explicitly, it is now skipped during the automatic generation step.
When the current CLDR data is applied, however, the result is as shown below.
Should alignment with CLDR be preferred?

```
    "#in {num}d": "#‏+{num} d",
    "#in {num}h": "#‏+{num} h",
    "#in {num}mi": "#‏+{num} min",
    "#in {num}mo": "#‏+{num} m",
    "#in {num}s": "#‏+{num} s",
    "#in {num}w": "#‏+{num} w",
    "#in {num}y": "#‏+{num} y",
    ...
```

tools/cldr/datefmts.js
seonmiJin: The changes have been applied as suggested, and it has been verified that this update does not result in any data changes.

tools/cldr/gendatefmts2.js
ehoogerbeets: No let's stay with what we have before. That CLDR data is placeholder data until they get the real data.

js/test/durfmt/testdurfmtasync.js
ehoogerbeets: That's very weird. I'm pretty sure that was a misvote in the voting tool... sometimes changes proposed by one person get only one or two votes and it is not enough to be statistically significant, so they just go with the proposal when it doesn't really make any sense.

Anyways, we'll just go with it. It's CLDR problem, not ours.

ehoogerbeets (approved):

PR #498, opened 2026-04-27
Update `assemble.mjs` to support the `--splitByLocale` option

gouniLee:
### Checklist

* Update `assemble.mjs` to support the `--splitByLocale` option. When enabled, locale data is written as a hierarchy of separate files (`root.json`,  `en.json`, `en-US.json`, etc.) instead of one fully-merged file per locale.  
Each file contains only the data for that specific sublocale level, allowing consumers to load exactly the layers they need. The `--splitByLocale` flag  requires `--mergeJson`; a warning is printed if used without it.

* Related PR to lib-mono : https://github.com/iLib-js/ilib-mono/pull/276
### Resolution

### Links


package.json
```diff
@@ -102,10 +102,14 @@
     "scripts": {
         "test": "ant test.dynamic.uncompiled",
         "build.web": "webpack -p",
+        "clean": "rm -rf assembled",
         "watch": "webpack --watch",
-        "build:mergeJson-compressed": "ilib-assemble --mergeJson --ilibincPath ./js/assembleData/ilib-inc.js --localefile ./js/assembleData/locales.json --customLocalePath ./js/assembleData/customData --compressed assembled",
-        "build:mergeJson": "ilib-assemble --mergeJson --ilibincPath ./js/assembleData/ilib-inc.js --localefile ./js/assembleData/locales.json --customLocalePath ./js/assembleData/customData assembled",
-        "debug:mergeJson": "node --inspect-brk node_modules/.bin/ilib-assemble --mergeJson --ilibincPath ./js/assembleData/ilib-inc.js --localefile ./js/assembleData/locales.json --customLocalePath ./js/assembleData/customData assembled"
+        "build:mergeJson-compressed": "ilib-assemble --mergeJson --ilibincPath ./js/assembleData/ilib-inc.js --localefile ./js/assembleData/locales.json --compressed assembled",
```
ehoogerbeets, line 107: The --customLocalePath is removed from these. Is that intentional? Maybe describe what that does and why it changed.

js/assembleData/assemble.mjs
ehoogerbeets, line 59: splitByLocale is not documented here

ehoogerbeets, line 30: Maybe add description of splitting by whole locales versus splitting by the locale pieces

package.json
gouniLee: Yes. It was intentionally removed. Since there is no directory corresponding to `customLocaleData` in this repository, the option is not valid here. Therefore, it has been removed.

js/assembleData/assemble.mjs
gouniLee: I updated it.

gouniLee: I'veㄷ added a more detailed description

ehoogerbeets (approved):

PR #501, opened 2026-05-08
Add in documentation for AI and for humans

ehoogerbeets:
### Checklist

### Resolution

### Links

### DIFF
--- a/.cursor/rules/development-process.mdc
+++ b/.cursor/rules/development-process.mdc
@@ -1,0 +1,37 @@
+---
+description: TDD workflow, approval gates, and test-first fixes for iLib
+alwaysApply: true
+---
+
+# Development process (agent)
+
+Follow **test-driven development** unless the user explicitly directs otherwise.
+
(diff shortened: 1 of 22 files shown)


.cursor/rules/development-process.mdc
```diff
@@ -0,0 +1,37 @@
+---
+description: TDD workflow, approval gates, and test-first fixes for iLib
```
gouniLee, line 2: These files are under .cursor/ and are intended for CursorAI. Each AI reads different configuration files, for example, if we want to additionally write specifications for Claude, I would probably need to create a claude.md file and specify that it should read the files under .cursor/.

gouniLee (approved):

.cursor/rules/development-process.mdc
ehoogerbeets: Yes, I think so. Also, cursor can use Claude as the model it chooses for doing its work.

PR #504, opened 2026-06-09
Resolve paths in `assemble.mjs` from module dir 

seonmiJin:
### Checklist

* [ ] This PR has passed all test cases on `QT/QML`.

### Summary
[//]: # (One or two sentences: what does this PR do and why?)
* Change path resolution from process.cwd() to module directory-based paths to fix issues when running from different working directories.

### Changes
[//]: # (Bullet list of the key files/functions changed and what was done)
* js/assembleData/assemble.mjs
  * Replace process.cwd() with import.meta.url and fileURLToPath to resolve paths from module location
  * Add validateRequiredIlibPaths() function to validate required iLib paths
* Related: https://github.com/iLib-js/ilib-mono/pull/306

### DIFF
--- a/js/assembleData/assemble.mjs
+++ b/js/assembleData/assemble.mjs
@@ -27,18 +27,10 @@ const reDependentPattern = /require\(\s*["']\.*\/([^"']+\.js)["']\);/g;
 const reDataPattern = /\/\/\s*!data\s+([^\n\r]+)/g;
 const moduleDir = path.dirname(fileURLToPath(import.meta.url));
 const ilibRoot = path.resolve(moduleDir, '..', '..');
-const ilibLibPath = path.join(ilibRoot, 'js/lib');
-const ilibLocaleDataPath = path.join(ilibRoot, 'js/data/locale');
 
-function validateRequiredIlibPaths() {
-    if (!existsSync(ilibLibPath)) {
-        throw new Error(`iLib JS library directory not found: ${ilibLibPath}`);
-    }
-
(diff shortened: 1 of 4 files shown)


js/assembleData/assemble.mjs
```diff
@@ -60,9 +75,11 @@ const reDataPattern = /\/\/\s*!data\s+([^\n\r]+)/g;
  * @returns {object} Merged JSON data map keyed by locale (or by sublocale path when splitByLocale is true)
  */
 export function assemble(ilibFiles, options) {
+    validateRequiredIlibPaths();
```
gouniLee: I think the `ilibPath` value passed through the ilib-assemble options could be validated using `options.opt?.ilibPath`. Is there a specific reason for checking the path again using moduleDir? `(path.dirname(fileURLToPath(import.meta.url));`

seonmiJin: I updated the logic to first reference `options.opt?.ilibPath`, and if it’s not provided, fall back to resolving `js/lib` and `js/data/locale` relative to the module directory.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "ilib",
-    "version": "14.22.0",
+    "version": "14.22.1",
```
gouniLee, line 3: `build.properties` file should be updated as well whenever the version is bumped.

seonmiJin: I've updated build.properties.

ehoogerbeets (approved):

PR #507, opened 2026-07-03
Include assemble scripts in publish and support package-root layout

gouniLee:
### Checklist

### Summary
Bundle assemble.mjs and assembleZoneinfoData.mjs into the published
npm package and make them work when run from the package root, not
just from the source tree.

- `build.xml`: copy `assemble.mjs` and `assembleZoneinfoData.mjs` into the
  export package directory
- `package.json.template`: add both files to the "files" list so they
  are included on npm publish
- `assemble.mjs`: resolve the lib/locale-data directories for both the
  source tree (js/lib, js/data/locale) and the published package
  (lib, locale); apply the same resolution to the ilibPath override
- related PR: https://github.com/iLib-js/ilib-mono/pull/315

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -21,6 +21,14 @@ Breaking Changes:
   Applications using iLib on Qt/QML platforms will need to pin to a previous
   version of iLib.
 
+New Features:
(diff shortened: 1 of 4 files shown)


ehoogerbeets (approved):
Just quickly check using tar that the mjs files exist in the ilib.tgz file at the root of the package so we know they are published.

PR #509, opened 2026-07-22
Extract PluralUtils from IString.js and fix plural rule evaluation bugs (Phase 1)

gouniLee:
### Checklist

### Summary

Extracts the plural rule engine (`plurals_default`, `loadPlurals`, `_fncs`) out of
`IString.js` into a standalone `PluralUtils.js` module (Phase 1 of the IString
refactoring plan), then refactors and fixes several correctness bugs surfaced while
adding unit tests.

`IString.js` shrinks by ~383 lines; the extracted module is 378 lines. **The public
API is unchanged** — `IString.loadPlurals`, `IString._fncs`, and
`IString.plurals_default` are retained as backward-compatible aliases, so none of the
38 `require('./IString')` callers need to change.

### What changed

**Refactor**
- Move the plural rule engine into `js/lib/PluralUtils.js`; `IString` keeps aliases for
  all three static members.
- Internal cleanup of the engine: local `fn` alias (removes 27 repetitions of the
  qualified name), inline `matchRange` into its callers, remove dead code
  (`firstProp`), `Array.isArray` instead of `Object.prototype.toString.call`, drop
  unused `ilib.bind`. Net −131 lines in the engine itself.

**Bug fixes** (all predate the extraction; all only affect the fallback path used when
`Intl.PluralRules` is unavailable, e.g. Node 10 with small-icu)
- `calculateNumberDigits`: use the **absolute value** so CLDR operands are correct for
  negative numbers; read the **full exponent** rather than just its first digit (e.g.
  `1e-21`); keep large integers intact instead of mangling them through `parseInt`;
  strip trailing zeros so the `w`/`t` operands are no longer identical to `v`/`f`.
- `matchRangeContinuous` / `inrange`: treat a 3+ value list (e.g. `[11,71,91]`) as a
  **set**, not a range; handle nested/mixed range lists (e.g. `[[3,10],[13,19]]`,
  `[[3,4],9]`).
- `neq`: guard `typeof(n) === 'object'` for symmetry with `eq`.

**Known limitation (documented, not fixed here)**
- The CLDR data encoding can't distinguish a two-value range `2..4` from a two-value
  set `2,4` — both collapse to `[2,4]`. The runtime assumes "range" (correct for e.g.
  Russian, wrong for Scottish Gaelic). Documented in `PluralUtils.js` with the root
  cause in `tools/cldr/genplurals.js`. A proper fix requires an unambiguous data format
  and regenerating every locale's `plurals.json`.

### Tests
- New `js/test/root/testpluralutils.js` — ~63 tests / 138 assertions covering
  `calculateNumberDigits` (negative, exponent, large-int), set-vs-range matching,
  nested/mixed lists, and several real locales (Breton, Russian, English). Passes in
  all three build modes (dynamic, dynamicdata, assembled).
- Verified against `Intl.PluralRules` for **214/215 locales** over 0..200 (only
  Scottish Gaelic differs — the documented data-format limitation).
- Updated `testunitfmt_be_BY` / `testunitfmt_mt_MT` temperature-2 expectations for the
  non-Intl branch, since the absolute-value fix changes the plural category of negative
  temperatures on that path only (be-BY → "many" `градусаў`; mt-MT → "many" `-il grad`).
  Intl-path expectations unchanged.

### Notes for reviewers
- This is Phase 1 only; Phase 2 (extracting `IStringFmt.js`) is planned separately and
  tracked in the refactoring doc.
- The bug fixes are CLDR-correct and confined to the iLib fallback path — behavior under
  modern Node/browsers (Intl path) is unaffected.


js/test/units/testunitfmt_be_BY.js
```diff
@@ -104,10 +107,10 @@ module.exports.testunitfmt_be_BY = {
             if (cldrVersion !== undefined && cldrVersion >= 36) {
                 test.equal(str, "-16,666666666666668 градуса Цэльсія");
             } else {
-                test.equal(str, "-16,666666666666668 градусы Цэльсія");
+                test.equal(str, "-16,666666666666668 градусаў Цэльсія");
```
ehoogerbeets, line 110: Why did this change? All you did is move the code to a new file, so the results should not be any different.

js/test/units/testunitfmt_mt_MT.js
```diff
@@ -88,7 +88,27 @@ module.exports.testunitfmt_mt_MT = {
             length: "long"
         });
         var str = uf.format(m1);
-        test.equal(str, "-16.666666666666668 grad Celsius");
+        var platform = ilib._getPlatform();
```
ehoogerbeets, line 91: Same here

js/test/units/testunitfmt_be_BY.js
gouniLee: The changes in this branch include not only splitting the PluralUtil.js file, but also improving the existing logic. In the process, a bug was also identified and fixed. The same conditional branch also exists in the browser case of the same test.

To explain this part in more detail:
`PluralUtils.calculateNumberDigits` was fixed to use Math.abs(number) before computing CLDR operands, as the spec defines operand n on the absolute value. With -16.666, the old code computed n mod 10 = -6 (no rule matched → fell through to few); the fixed code computes n mod 10 = 6, which falls in the 5..9 range → many → градусаў.

This only affects the non-Intl.PluralRules path (Node 10 / small-icu). Modern Node uses Intl.PluralRules directly and is unaffected.

js/test/units/testunitfmt_mt_MT.js
gouniLee: This is for the same reason I mentioned in the comment above.

js/lib/PluralUtils.js
```diff
@@ -0,0 +1,378 @@
...
+        }
+
+        // Derive the fraction operands from the text after the decimal point.
+        // Two known inaccuracies here, both only affecting values that never
+        // occur in real plural contexts (see the limitation note above):
+        //   - w always equals v and t always equals f, because a JS number's
+        //     string form never carries trailing zeros (1.50 -> "1.5").
+        //   - for very small/large decimals JS uses exponential notation
+        //     (1.5e-9 -> "1.5e-9"), so decStr becomes "5e-9" and v/f are wrong.
+        //     Correct handling would require expanding the exponent first.
```
ehoogerbeets, line 200: Can we put "TO DO" here to make sure we fix this bug in the future?

ehoogerbeets (approved):

js/lib/PluralUtils.js
gouniLee: I now remember why this part wasn't left as a TODO to be fixed, but was instead documented as a note describing the cases where it misbehaves. because it wasn't clear whether supporting these cases was really the right thing to do.

Here's a more detailed explanation:
These extreme edge cases (values like 1.5e-9, 1e21 that JS renders in exponential notation) are very unlikely to occur in real plural-rule evaluation, and fixing them properly would require changing the function's input contract — from a plain number to a formatted string — which would ripple out to every call site, not just this function. For that reason, this limitation was deliberately left unfixed and documented as a comment instead of being addressed in code.
Specifically:
* `calculateNumberDigits` is only a fallback path for environments without `Intl.PluralRules` (see the docstring above the function, lines 163–169). Modern runtimes never hit this code.
* Plural rule selection in practice essentially never needs to distinguish values at this magnitude (1.5e-9, 1e21, etc.).

Current implementation status
```
var n = 1.5e-9;
n.toString();        // "1.5e-9"
// dotIdx = 1, so decStr = "5e-9"
// v = decStr.length = 4   (should be 9 fraction digits)
// f = parseInt("5e-9") = 5  (should be 15)
```

PR #510, opened 2026-08-06
Change the name from "Nauru" to "Naoero"

ehoogerbeets:
- following the government's request to all other nations and a constitutional amendment to follow the new naming which more closely aligns with the way that you say the name of the country in their native language

### Checklist

* [x] `ReleaseNotes` has been updated or is not needed.
* [x] This PR has passed all test cases on `Node.js` and `Chrome Browser`.

### Summary
[//]: # (One or two sentences: what does this PR do and why?)


docs/demo/scripts/resources.js
```diff
@@ -119,7 +119,7 @@ ilib.data.ilibTutorial = {
 	"en-NA": "English - Namibia",
 	"en-NF": "English - Norfolk Island",
 	"en-NG": "English - Nigeria",
-	"en-NR": "English - Nauru",
+	"en-NR": "English - Naoero",
```
gouniLee, line 122: Until this change is reflected in the CLDR data, it also needs to be applied to the CLDR Tool code (i.e., gencountrynames.js) so that it will not be overwritten again during the next CLDR update.

Also, it would be good to add country test code for the related data changes.

seonmiJin (approved):

js/data/phone/ilib-phoneres.xliff
```diff
@@ -835,8 +835,8 @@
             <target>Kuba</target>
          </trans-unit>
          <trans-unit id="77576">
-            <source>Nauru</source>
-            <target>Nauru</target>
+            <source>Naoero</source>
```
gouniLee, line 838: note. I checked that `ilib-phoneres.xliff` is a required file for the currently invalid `novaloc`, and that it is called from `genphone.sh`. I think this part will need to be updated in the future so that it can work with the current version.

js/data/locale/uz/Arab/ctrynames.json
```diff
@@ -168,7 +168,7 @@
     "Muqaddas Yelena oroli": "SH",
     "Myanma (Birma)": "MM",
     "Namibiya": "NA",
-    "Nauru": "NR",
+    "Naoero": "NR",
```
gouniLee, line 171: What about the unit tests for these changed country data?

PR #511, opened 2026-08-07
Extract IStringFmt.js from IString.js (Phase 2)

gouniLee:
### Checklist

## Summary

Phase 2 of the IString.js split described in `docs/IString-Refactoring.md`. Extracts the formatting-method cluster (`format`, `_testChoice`, `_isIntlPluralAvailable`, `formatChoice`, `setLocale`, `getLocale`) out of `IString.js` into a new `IStringFmt.js`, mixed back onto `IString.prototype` via `Object.assign`. No change to the public API.

- Move the six formatting methods into `js/lib/IStringFmt.js` as a plain object (mirrors the `PluralUtils.js` pattern from Phase 1), required and mixed into `IString.prototype` from `IString.js`
- `formatChoice`'s internal `new IString(...)` calls become `new this.constructor(...)` so `IStringFmt.js` never requires `IString.js` (no circular dependency)
- Restore `IString.prototype.constructor = IString`, which is lost when `.prototype` is replaced with an object literal — required for the `this.constructor` calls above to resolve correctly
- `_testChoice`/`setLocale` call `PluralUtils` directly instead of going through the `IString._fncs`/`IString.loadPlurals` static aliases
- Add `js/test/root/teststringfmt.js` and register it in `testSuiteFiles.js`

`IString.js`: 1158 → 712 lines. `IStringFmt.js`: 482 lines (new).

## Notes

This is a pure move — no behavior changes. Several pre-existing logic bugs in the moved `formatChoice`/`format` code were found during review (not caused by this extraction, just carried over verbatim) and are being fixed separately in the `fix-istringfmt-logic-bugs` branch, to keep the structural move and the behavioral fixes easy to review independently.

This PR's base branch is `refactor-istring-plural-utils`, which is #509. (Phase 1 — extract `PluralUtils.js`) needs to be merged first; this PR should be rebased onto `development` (or its base updated) once that lands.

## Test plan

- [x] `ant test.root` — 7092 assertions, all passing
- [x] `ant test.address` / `test.date` / `test.units` / `test.name` — consumer suites (`formatChoice`/`format`/`setLocale` callers: `DateFmt`, `DurationFmt`, `UnitFmt`, `ResBundle`, `NameFmt`, `AddressFmt`) all passing, no regressions
- [x] `docs/IString-Refactoring.md` updated to mark Phase 2 done


js/lib/IString.js
```diff
@@ -171,418 +171,6 @@ IString.prototype = {
...
-     * or an array of indices
-     * @param {Object} params The hash of parameter values that replace the replacement
-     * variables in the string
-     * * @param {boolean} useIntlPlural [optional] true if you are willing to use Intl.PluralRules object
-     * If it is omitted, the default value is true
-     * @throws "syntax error in choice format pattern: " if there is a syntax error
-     * @return {string} the formatted string
-     */
-    formatChoice: function(argIndex, params, useIntlPlural) {
-        var choices = this.str.split("|");
```
ehoogerbeets, line 476: Please keep this method and its documentation here, but delegate to the IStringFmt code instead.

```
formatChoice: function(argIndex, params, useIntlPlural) {
    return IStringFmt.formatChoice(argIndex, params, useIntlPlural);
}
```
Same for the other IString methods

js/lib/IStringFmt.js
```diff
@@ -0,0 +1,482 @@
...
+ * easily have been written as ",one" instead of "other,one" and the result will be the same.
+ *
+ * @param {*|Array.<*>} argIndex The index into the choice array of the current parameter,
+ * or an array of indices
+ * @param {Object} params The hash of parameter values that replace the replacement
+ * variables in the string
+ * * @param {boolean} useIntlPlural [optional] true if you are willing to use Intl.PluralRules object
+ * If it is omitted, the default value is true
+ * @throws "syntax error in choice format pattern: " if there is a syntax error
+ * @return {string} the formatted string
```
ehoogerbeets: These should be @private methods. Leave the documentation in IString instead.

js/lib/IString.js
gouniLee: I've updated the code and documentation overall in relation to this. Please review them again.

js/lib/IStringFmt.js
gouniLee: Yes. I marked all the methods in IStringFmt as private.

js/lib/IString.js
```diff
@@ -171,418 +171,6 @@ IString.prototype = {
...
-     * <pre>
-     * There are 12 objects in the {container}.
-     * </pre>
-     *
-     * The result can then be formatted again with a different parameter block that
-     * specifies a value for the container property.
-     *
-     * @param params a Javascript object containing values for the replacement
-     * parameters in the current string
-     * @return a new IString instance with as many replacement parameters filled
```
ehoogerbeets, line 229: This jsdocs + method (format) are still deleted from this file.

js/lib/IString.js
ehoogerbeets: The formatChoice() method and documentation is still missing in this PR... did you forget to push maybe?

gouniLee: it seems this is because the lines shifted due to the update. In the updated file, `format()` is at line 762, and `formatChoice()` is at line 904. Without missing any JSDoc.I'll try to match the line numbers as closely as possible.

gouniLee: I updated it to match the original methods' positions.This makes the diff much easier to see.

js/lib/IString.js
```diff
@@ -1155,4 +936,9 @@ IString.prototype = {
     }
 };
 
+// restore the constructor reference lost when the prototype was replaced with
+// an object literal above; IStringFmt.formatChoice relies on this.constructor
+// to create new IString instances without requiring IString.js itself
+IString.prototype.constructor = IString;
```
ehoogerbeets: Since we are not doing a mixin any more, we don't need this one still... yes?

gouniLee: This line is needed for the following two reasons: 

Cause 1 — Prototype replaced wholesale

IString.prototype = {...}   // line 154
Overwriting .prototype with a new object literal drops the auto-attached constructor property; it falls back to Object. This predates the refactor.

Cause 2 — formatChoice relies on this.constructor

// IStringFmt.js, lines 196, 199, 222, 230, 240, 247
result = new this.constructor(strings[idx]);
Once formatChoice moved into IStringFmt.js, using new IString(...) directly would create a circular require (IStringFmt.js → IString.js → IStringFmt.js). new this.constructor(...) avoids that. 
(i.e original one [code](https://github.com/iLib-js/iLib/blob/development/js/lib/IString.js#L904): )

Combined effect: Cause 1 breaks this.constructor; Cause 2 depends on it. Line 942 (IString.prototype.constructor = IString;) restores the link so formatChoice builds IString instances instead of plain Objects.

js/lib/IString.js
ehoogerbeets:
```suggestion
Object.assign(IString.prototype, {
```

js/lib/IString.js
ehoogerbeets: See next comment

gouniLee:
@ehoogerbeets It looks like you haven't submitted the review comments you left earlier. I saw what you were writing before, but I can't see it now. I've updated the code as discussed in the meeting. Please check it again.

ehoogerbeets (approved):

